PR #35, opened 2021-01-07
add Macro syntax for argmax and argmin

### LINKED ISSUE #31: [lecture_comparison][rational_expectations] \argmax Issue
Dear @mmcky , I find another issue:
- ``\argmax`` is not compiled in an equation of lecture [rational_expectations](https://quantecon.github.io/lecture-python.myst/rational_expectations.html)

Please see the following (left: ``MYST``, right: ``RST``)
Screen Shot 2021-01-06 at 3 58 42 pm

mmcky:
This PR adds TeX macros for `argmin` and `argmax` which should fix #31


mmcky:
@shlff we should also fix other cases

Would you have time to go through:

https://github.com/QuantEcon/lecture-python/blob/b37408c7b5aeb3875767949c6449113bcd4b1702/conf.py#L263

and let me know which ones we need?

I would suggest using `grep` to see if they are needed. For example

```bash
grep -r "RR" source/rst/*
```

will give you a list of all cases of `RR`

shlff:
> @shlff we should also fix other cases
> 
> Would you have time to go through:
> 
> https://github.com/QuantEcon/lecture-python/blob/b37408c7b5aeb3875767949c6449113bcd4b1702/conf.py#L263
> 
> and let me know which ones we need?
> 
> I would suggest using `grep` to see if they are needed. For example
> 
> ```shell
> grep -r "RR" source/rst/*
> ```
> 
> will give you a list of all cases of `RR`

Thanks @mmcky . I checked all cases from ``\RR`` to ``\argmin`` in the [list](https://github.com/QuantEcon/lecture-python/blob/b37408c7b5aeb3875767949c6449113bcd4b1702/conf.py#L263) through all ``lecture-python`` lectures via ``grep`` and find 
- ``\argmax`` is the only case we need to fix

Do you want me to check more cases? If not, then I think this PR can be merged safely.

AakashGfude:
LaTeX equivalent of this is done in https://github.com/QuantEcon/lecture-python.myst/pull/35

PR #101, opened 2021-02-11
ENH: enable cached build folders using artifacts

### LINKED ISSUE #12: [github actions] Look at how to make builds quicker using cache
Perhaps the best strategy here is to:

1. use `publish.yml` to build a fresh build and upload `_build` as an artifact
2. PR's use that _build folder (from a single source) to speed up builds by downloading the artifact

mmcky:
This PR fixes #12 by using `build` artifacts from the `publish.yml` process. 

- [x] add use of `cached build` folder for PRs
- [ ] add use of `cached build` folder for publishing with `_build` folder updating

Not going to implement the second as it is more risky given it is on the `publishing` pathway. We can add this at a later date.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
           workflow: publish.yml
           branch: main
           name: build-cache
+          path: _build
       - name: Build HTML
         shell: bash -l {0}
         run: |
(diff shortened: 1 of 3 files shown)


mmcky:
artifacts seem still to be limited to individual `workflows` and passing between `jobs` rather than passing between `workflows` -- will revert to use `cache`

mmcky:
Until Github Actions supports this we should use: https://github.com/dawidd6/action-download-artifact

The issue with `cache` is in the regular updating of the `_build` folder. 

This way PR's can use the latest `_build` artifacts from the `publish.yml` file.

Issue #102, opened 2021-02-12
Revert changes to `align` on `main` branch


mmcky:
@AakashGfude I think we need to revert https://github.com/QuantEcon/lecture-python.myst/commit/6a2d1e07695159b4954ca3041ee52cb1d695d5b9 (and any associated commits) and update these in the `rst` source files. 

`align` -> `aligned` (as aligned is not a equation environment and can be inside `equation`)

https://quantecon.github.io/QuantEcon.manual/writing/styleguide.html#mathematical-notation

mmcky:
- [ ] coordinate a search for `\begin{align}` on lecture-python and setup a PR to update the sources

PR #103, opened 2021-02-12
🐛 FIX: Math env syntax issue

AakashGfude:
fixes https://github.com/QuantEcon/lecture-python.myst/issues/102

@mmcky only a few changes needed. rest seems ok.

### DIFF
--- a/lectures/exchangeable.md
+++ b/lectures/exchangeable.md
@@ -116,10 +116,10 @@ $$
 Using the laws of probability, we can always factor such a joint density into a product of conditional densities:
 
 $$
-\begin{align}
+\begin{aligned}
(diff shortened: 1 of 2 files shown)


mmcky:
thanks @AakashGfude.

I don't see where #70 is reverted. Am I missing something?

AakashGfude:
> I don't see where #70 is reverted. Am I missing something? We should implement those changes in the `rst` repo.

#70 did the same thing as this PR, but with different edits. I think you are right, we should correct the rst sources instead of correcting it here.

mmcky:
oh ok -- so that was reverted on `main` thanks @AakashGfude. Do you mind setting up the PR on lecture-python when you get the chance and I will merge those changes in.

mmcky:
This should not longer be needed with the lastest updates on `main`

PR #132, opened 2021-04-19
MAINT: revert back to pip installs using quantecon-book-theme=0.1.7

mmcky:
`quantecon-book-theme=0.1.7` has now been released on PyPI

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - pip:
     - jupyter-book
     - sphinx-multitoc-numbering
-    - git+https://github.com/quantecon/quantecon-book-theme
+    - quantecon-book-theme
     - sphinx-tojupyter
     - sphinxext-rediraffe
     - sphinx-exercise
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - pip:
     - jupyter-book
     - sphinx-multitoc-numbering
-    - quantecon-book-theme=0.1.6
+    - quantecon-book-theme==0.1.6
     - sphinx-tojupyter
     - sphinxext-rediraffe
     - sphinx-exercise
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - pip:
     - jupyter-book
     - sphinx-multitoc-numbering
-    - quantecon-book-theme==0.1.6
+    - quantecon-book-theme
     - sphinx-tojupyter
     - sphinxext-rediraffe
     - sphinx-exercise
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - pip:
     - jupyter-book
     - sphinx-multitoc-numbering
-    - quantecon-book-theme
+    - quantecon-book-theme=0.1.6
     - sphinx-tojupyter
     - sphinxext-rediraffe
     - sphinx-exercise


mmcky:
@AakashGfude the latest release of `quantecon-book-theme` contained the fixes for math. So I am not sure why this test build is showing the inline math colliding more than the current live site:

The sentence below this section title demonstrates the issue. 

https://607ceb2f1d8fad05445fac10--nostalgic-wright-5fa355.netlify.app/linear_algebra.html#id7

VS

https://python.quantecon.org/linear_algebra.html#id7

Any ideas?

mmcky:
looks like 

```
quantecon-book-theme               0.1.5
```

is installed

mmcky:
This is from: https://github.com/QuantEcon/lecture-python-advanced.myst/pull/4#issuecomment-824521646

@mmcky @DrDrij seems like the math fix itself is actually causing the issue https://github.com/QuantEcon/quantecon-book-theme/pull/96/files. When I remove the css styling here https://607ceb2f1d8fad05445fac10--nostalgic-wright-5fa355.netlify.app/linear_algebra.html#vectors in safari, the "or" does not overlap.

Maybe we need to do a more holistic test, before deciding on a fix.

mmcky:
@AakashGfude what I am confused about is that the current `live` site uses the master branch:

git+https://github.com/quantecon/quantecon-book-theme

while this PR is just switching to `quantecon-book-theme=0.1.5` which should be the same right?

AakashGfude:
> @AakashGfude what I am confused about is that the current `live` site uses the master branch:
> 
> git+https://github.com/quantecon/quantecon-book-theme
> 
> while this PR is just switching to `quantecon-book-theme=0.1.5` which should be the same right?

@mmcky that is true. looks like the current live site, does not have that CSS change, from what I can see. After the 0.15 release, has the production site been built and deployed again?

mmcky:
@AakashGfude so there is definitely a difference between installing from `master` directly and installing via `pip`. I just did a test where I setup `quantecon` environment using the `main` branch and the `html` looks OK, and then I setup a `quantecon` environment from this branch and the math is not OK.

mmcky:
So when comparing the `site-packages` between:

1. installing from `main`
2. installing from `pip`

there is one significant difference

```
▶ diff -r quantecon_book_theme-main quantecon_book_theme-branch
Binary files quantecon_book_theme-main/__pycache__/__init__.cpython-38.pyc and quantecon_book_theme-branch/__pycache__/__init__.cpython-38.pyc differ
Binary files quantecon_book_theme-main/__pycache__/launch.cpython-38.pyc and quantecon_book_theme-branch/__pycache__/launch.cpython-38.pyc differ
Only in quantecon_book_theme-branch/static: quantecon-book-theme.css
```

Here the `quantecon-book-theme-branch` is being installed from `pip`. So it looks like this `css` file is only getting included `via pip`.

mmcky:
Looking at the `css` files the `preview` of this branch includes a `css` file that **doesn't** have these math changes

https://github.com/QuantEcon/quantecon-book-theme/commit/b425870af28c6869eed1779269f28252428d64f5

while `gh-pages` does (which installs directly from `github@main`)

mmcky:
> When I remove the css styling here https://607ceb2f1d8fad05445fac10--nostalgic-wright-5fa355.netlify.app/linear_algebra.html#vectors in safari, the "or" does not overlap.

So the `css` styling should be removed and `pip` has been released with an old version of the `css` file. (https://github.com/QuantEcon/quantecon-book-theme/issues/106)

mmcky:
This PR will now update to use `quantecon-book-theme=latest` which will be `0.1.6`

mmcky:
thanks @AakashGfude  for releasing `quantecon-book-theme=0.1.6` looks pretty good -- but there seems to be one element missing which is the right hand side local toc element. 

Screen Shot 2021-04-30 at 9 17 57 pm

mmcky:
@AakashGfude going to merge this in so that `0.1.7` is new `baseline` then will open a test for upcoming `0.2.0`

PR #141, opened 2021-05-17
[intro] wrap intro sentence about website into an only directive for html

mmcky:
This will remove the sentence from non-html builds.


mmcky:
@jstac this gets rid of:

1. website sentence mentioned in your email
2. the additional `tableofcontents` directive listing

PR #158, opened 2021-06-28
[short_path] Define num_nodes in compute_to_go

tetrapod54:
`num_nodes`  in the `compute_cost_to_go` function refers to that in the global scope, but it should be defined in the local scope.


jstac:
True!  Thanks @tetrapod54 , well spotted.

PR #159, opened 2021-06-28
[short_path] Delete next_J in compute_cost_to_go

Eri-Oyama:
Remove redundant array creation.


jstac:
Nice work @Eri-Oyama .  Thank you.

PR #162, opened 2021-06-29
[FIX] Fixes short_path image issues

mmcky:
The issue was that `jupyter-cache` is executing **all** notebooks in the source repository, including those in `_static/lecture_specific` which was rebuilding the images for `short_path` via the notebook included to build them. However `pylab` was used to show the figures in the notebook and the software must have changed to clear the `matplotlib.fig` object after showing it in the notebook causing the images to be written as empty images.

cc: @AakashGfude @jstac


mmcky:
**Note:** *this issue won't come up locally unless you fully clear the cache and have the latest environment.

PR #172, opened 2021-07-19
[career] Correct variable in gen_path

kuramogura:
The current code happens to works, because `greedy_star` refers to that in the global scope.


jstac:
Thanks @kuramogura , good catch!

PR #179, opened 2021-08-18
[mathjax] enable chtml since v3 updated to v3.2

mmcky:
See if `chtml` can be used instead of `svg` for math rendering


mmcky:
@shlff @AakashGfude this `chtml` looks really nice now -- if you had 10 minutes would you mind taking a quick look at a few pages and let me know what you think.

mmcky:
This is looking pretty good -- even handles the long equations in

https://611c65f7767a946576581014--nostalgic-wright-5fa355.netlify.app/re_with_feedback.html#fun-with-sympy-code

well.

mmcky:
@AakashGfude I think I figure it out with web inspector

```
e.mathjax = {
                    version: "3.2.0",
```

so `3.2` is loading 👍

AakashGfude:
> @AakashGfude I think I figure it out with web inspector
> 
> ```
> e.mathjax = {
>                     version: "3.2.0",
> ```
> 
> so `3.2` is loading 👍

Nice work @mmcky .

shlff:
> This is looking pretty good -- even handles the long equations in
> 
> https://611c65f7767a946576581014--nostalgic-wright-5fa355.netlify.app/re_with_feedback.html#fun-with-sympy-code
> 
> well.

Thanks @mmcky . They look nice to me. 

I guess this long equation is a bit so long that we could also modify the code a bit.

Additionally I notice that the horizontal lines in a fraction and a square root are missing. Will we fix that as well? e.g., the following:
Screen Shot 2021-08-18 at 3 10 27 pm

mmcky:
@shlff nice work! I didn't see the missing horizontal lines but that is a problem still!

https://python.quantecon.org/re_with_feedback.html#fun-with-sympy-code

Might need to investigate this before we switch to `chtml`

mmcky:
@AakashGfude @shlff 

https://61a40e5836e4295a847bfa2f--nostalgic-wright-5fa355.netlify.app/re_with_feedback.html

these are now working with `chtml` so we should be able to turn `svg` off soon. 

There are now a lot of `matplotlib` deprecation warnings though :-(

shlff:
Thanks @mmcky . Those warnings are raised because the following three functions were deprecated in Matplotlib 3.4 and will be removed soon in matplotlib:
- ``to_png``
- ``to_rgba``
- ``to_mask``
- ``MathtextBackendBitmap``

I will try to search for their substitutes and see if I can remove these warnings. I will keep you posted here later.

mmcky:
@jstac the font size for math in `chtml` seems a bit bigger than under `svg`. Interested in your thoughts for the inline math (in particular).

mmcky:
`chtml` has been enabled so this is no longer required. 

@shlff I have checked the test case you mentioned above and it looks nice now. 

https://641a5b9009dd68246c2dc87b--nostalgic-wright-5fa355.netlify.app/re_with_feedback.html#fun-with-sympy

PR #192, opened 2021-11-19
Update mccall_model.md

amirsssss:
there were two iterations on variable i in nested loops, I changed one to counter


jstac:
Good catch @amirsssss!  Many thanks.

PR #200, opened 2022-02-03
FIX Enable default image conversion to png for now

mmcky:


AakashGfude:
`<div class="highlight">` in the output stream divs, when set to `display:none` resolves the infinite scroll issue. Still figuring out, what property is actually causing this.

mmcky:
@AakashGfude I am going to merge this so we can publish the latest commits from @thomassargent30. But let's keep figuring this out and we can apply a fix when we get one.

PR #201, opened 2022-02-07
Tags in code-cell not under brackets

AakashGfude:


mmcky:
thanks @AakashGfude an alternative would be to use

````
```{code-cell}
tags: [hide-output]

<code>
```
````

right?

AakashGfude:
@mmcky there is a leading space to tags here, and in the documentation, it seems. But, maybe the CSS part of it is not bullet-proof.

AakashGfude:
will briefly look at what difference in HTML/CSS this PR change makes.

mmcky:
We will need to get rid of the the first `:` I think

AakashGfude:
@mmcky corrected it, sorry this PR won't help, as because of that first `:` the block didn't render, and so the page was of the correct length. But it does reinforce the fact that the additional height is coming from the hidden output. As the browser adds the height of the hidden output to the page. If you unhide the hidden output, the page height is correct.

AakashGfude:
Which is because we are using `visibility: hidden` attribute, to hide, show elements for the code cell output. which hides the element, but preserves its height in the page, we should be using `display: none`. 
As pointed out here:  https://www.w3schools.com/cssref/pr_class_visibility.asp
`Tip: Hidden elements take up space on the page. Use the display property to both hide and remove an element from the document layout!`

AakashGfude:
Even though it looks like a `visibility` property issue, it does work on other lectures.
Seems to be caused by `retrying` package in 
`!conda install -y -c plotly plotly plotly-orca retrying`
Somehow it is mimicking what `visibility` bug should look like

mmcky:
@AakashGfude can you remind me if we need this or not?

AakashGfude:
@mmcky we don't need this. It did not help with the scrolling issue on the deployed website. I will close this PR. 
I think that infinite scrolling is handled in the main website now?

PR #205, opened 2022-02-17
FIX: Jax and M1 Compatibility

mmcky:
A simple fix is to fetch `jax` from `conda-forge` which provides the `no arch` version of `jax` for `Apple Mac arm64 users`


mmcky:
Closing due to #214

Issue #207, opened 2022-02-27
A question about the lecture "Samuelson Multiplier Accelerator"


jxx87:
Hello,
In the lecture "Samuelson Multiplier Accelerator," a snippet of codes is copied below:

```
def plot_simulation(self, ts_length=100, stationary=True):
    # Set distribution parameters equal to their stationary values for simulation
    if stationary == True:
        try:
        self.μ_x, self.μ_y, self.σ_x, self.σ_y = self.stationary_distributions()
        self.μ_0 = self.μ_y
        self.Σ_0 = self.σ_y
      ...
    x, y = self.simulate(ts_length)
```
I wonder why:
```
self.μ_0 = self.μ_y
self.Σ_0 = self.σ_y
```
Instead of:
```
self.μ_0 = self.μ_x
self.Σ_0 = self.σ_x
```
Since my understanding is that when later using simulate(), x_0 first draws from N (μ_0, Σ_0). So shouldn't x_0 draws from N(μ_x, σ_x) instead of N(μ_y, σ_y)?

shizejin:
Hello,

Thank you very much for your spotting of this! You are correct, and we have fixed in a new PR #210, and there are some other changes that you could take a look if interested.

Thanks again!

PR #210, opened 2022-03-04
Modify `samuelson`

shizejin:
Closes https://github.com/QuantEcon/lecture-python.myst/issues/207

This PR corrects several minor errors in the simulation when using the `LinearStateSpace` class.

1. the attribute of the `LinearStateSpace` for controlling the initial distribution is `mu_0` instead of `μ_0`
2. to be less confusing, I changed all the attribute names related with `LinearStateSpace` to their names denoted in the [source code](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/lss.py#L67)
3. when setting the initial distribution to be the stationary distribution, we should use `mu_x` and `Sigma_x` instead of `mu_y` and `Sigma_y`
4. when the "stationary" mode is off, the code still unintentionally simulates data from its stationary distribution, as we were setting `μ_0` at the stationary values. I changed `μ_0` slightly so that the simulation shows a convergence to the stationary distribution from somewhere away

### DIFF
--- a/lectures/samuelson.md
+++ b/lectures/samuelson.md
@@ -1216,7 +1216,7 @@ G = [[γ + g, ρ1,   ρ2],         # this is Y_{t+1}
      [γ,      α,    0],         # this is C_{t+1}
      [0,      β,   -β]]         # this is I_{t+1}
 
-μ_0 = [1, 100, 100]
+μ_0 = [1, 100, 50]
 C = np.zeros((3,1))
 C[1] = σ # stochastic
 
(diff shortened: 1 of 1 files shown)


jstac:
Thanks @shizejin !  Does this close https://github.com/QuantEcon/lecture-python.myst/issues/207 ?

shizejin:
> Thanks @shizejin ! Does this close #207 ?

Hi @jstac, yes it does! I have added a linking to the PR to close the issue when merged.

thomassargent30:
To me this looks very good. Thanks Zejin.

PR #214, opened 2022-03-17
FIX: update for compatibility for jax>=0.3.2

### LINKED ISSUE #213: [back_prop] Issue with jax>=0.3.2
The release of `jax==0.3.2` last night has deprecated the `index_update` function.

https://github.com/google/jax/blob/main/CHANGELOG.md#jax-032-march-16-2022

The replacement syntax is to use

https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html

mmcky:
fixes #213

### DIFF
--- a/lectures/back_prop.md
+++ b/lectures/back_prop.md
@@ -16,7 +16,7 @@ kernelspec:
 ```{code-cell} ipython3
 :tags: [hide-output]
 
-!pip install jax jaxlib
+!conda install -y -c conda-forge jax jaxlib
 !conda install -y -c plotly plotly plotly-orca retrying
 ```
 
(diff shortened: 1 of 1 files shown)


lectures/back_prop.md
```diff
@@ -329,25 +328,25 @@ def compute_xδw_seq(params, x):
...
+        bs = bs.at[i].set(b[0])
+        xs = xs.at[i+1].set(activation)
 
     final_w, final_b = params[-1]
     preds = final_w * xs[-2] + final_b
     
     # Store elements
-    δ = index_update(δ, -1, 1.)
-    ws = index_update(ws, -1, final_w[0, 0])
-    bs = index_update(ws, -1, final_b[0])
```
mmcky, line 349: @thomassargent30 the line starting with `bs=` previously had `ws` in the index_update function but I think it should be `bs`. Is that correct?

PR #222, opened 2022-04-10
Fix old links

jlperla:


jlperla:
@mmcky  A few minor link updates

mmcky:
thanks @jlperla -- really appreciate the updates. I'll run a link checker over the lectures too.

PR #229, opened 2022-05-12
Edit typos and prettify code in util_rand_resp.rst

natashawatkins:
- @mmcky is there a way to cite equations in text? the equation labels in the text didn't match the automatic equation labs. I changed the references but guessed the syntax.
- I think an article stills need citation? "XXXX Ljungqvist"
- I removed the font selection because it was producing an error

### DIFF
--- a/lectures/util_rand_resp.md
+++ b/lectures/util_rand_resp.md
@@ -429,7 +429,7 @@ Properties of  iso-variance curves  are:
 
 - From $V_1$ to $V_9$, the variance of the iso-variance curve increase monotonically, as colors  brighten monotonically
 
-Suppose the parameters of the iso-variance model follow those in article XXXX Ljungqvist, which are:
+Suppose the parameters of the iso-variance model follow those in Ljungqvist {cite}`ljungqvist1993unified`, which are:
 
 - $\pi=0.3$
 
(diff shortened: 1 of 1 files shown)


mmcky:
Thanks @natashawatkins.

Yes you can either use a `label` in the `math directive` or you can use `(label)` syntax immediately after a `$$`

https://jupyterbook.org/en/stable/content/math.html#numbering-equations

lectures/util_rand_resp.md
```diff
@@ -173,19 +173,19 @@ $$
 \text{Pr}(\text{yes}|A^{'})\times \text{Pr}(A|\text{yes})
 $$ (eq:util-rand-eight-b)
 
-This measure is just the first term in $(7)$, i.e., the probability that an individual answers "yes" and is perceived to belong to A.
+This measure is just the first term in `ref:util-rand-seven-a`, i.e., the probability that an individual answers "yes" and is perceived to belong to $A$.
```
mmcky: This should be

```
{eq}`eq:util-rand-seven-a`
```

to use the `eq` reference role

mmcky: the `eq` that has been used in the actual `label` is pretty redundant so could also be removed.

natashawatkins:
@mmcky I updated the equation references, thanks

mmcky:
thanks @natashawatkins 

I guess preview doesn't get built as your forked repo doesn't have netlify credentials.

PR #235, opened 2022-06-22
FIX: Plot rendering in Samuelson Lecture

mmcky:
The string for one of the plots should be formed without any included newlines for mathjax to render `$` properly


mmcky:
@thomassargent30 this should fix that plot issue.

Issue #246, opened 2022-08-04
Missing arguments in McCall with Fitted Value Function Iteration


Okroshiashvili:
The **McCallModelContinuous** class constructor is missing μ and σ if we want to control distribution of wages inside the class. If we don't want that, then inside **update** method μ and σ are redundant and removing them will make code runnable and will not throw exception.

[Here is the lecture link](https://github.com/QuantEcon/lecture-python.myst/blob/main/lectures/mccall_fitted_vfi.md)

jstac:
@Okroshiashvili Many thanks for pointing that out.   @shlff, do you have time to review this and fix as necessary?

shlff:
Sure thanks @jstac and @Okroshiashvili . I've had a look at the code, and I agree with @Okroshiashvili .

Here is a PR that fixes the code without affecting the results of current remaining code.

jstac:
Fixed in #248

PR #248, opened 2022-08-07
[mccall_fitted_vfi] fix issue #246

shlff:
Hi @jstac , this PR fixes issue #246 while keeping the results from the remaining code cells the same for lecture ``mccall_fitted_vfi``.


jstac:
Many thanks @shlff , much appreciated.  Merging.

jstac:
Fixed in #248

PR #254, opened 2022-08-28
Fix Taylor series about g=0 in Geometric series lecture

### LINKED ISSUE #212: Error in Geometric series lecture?
Thanks to Richard Rivero for this:

I’m not sure you have the right Taylor series for G^{T+1} about g=0, somewhere down the page. 

https://python.quantecon.org/geom_series.html 

I would think it would be more something like:

image

Smit-create:
Fixes #212


jstac:
Many thanks @Smit-create

PR #255, opened 2022-08-30
[geom_series] fix MatplotlibDeprecationWarning

shlff:
Hi @jstac , this PR fixes the MatplotlibDeprecationWarning in lecture ``geom_series`` (left: before fixing; right: after fixing):
Screen Shot 2022-08-30 at 12 00 42 pm


mmcky:
thanks @shlff

PR #256, opened 2022-08-31
Few bug fixes on convergence conditions

### LINKED ISSUE #175: BUG: Correct convergence conditions in function iteration
See #171

This pattern, which is found in [several files](https://github.com/QuantEcon/lecture-python.myst/search?q=%22if+verbose+and+i+%3C+max_iter%22),

```py
    while i < max_iter and error > tol:

    ...

    if i == max_iter:
        print("Failed to converge!")

    if verbose and i < max_iter:
        print(f"\nConverged in {i} iterations.")
```

should be corrected to, e.g.,

```py
    if error > tol:
        print("Failed to converge!")
    elif verbose:
        print(f"\nConverged in {i} iterations.")
```

### LINKED ISSUE #176: BUG: Scope of index variable in `for` loop
Bug found by @tetrapod54:

In [this block](https://github.com/QuantEcon/lecture-python.myst/blame/main/lectures/mccall_model.md#L485-L493), the variable `i` in `for i in range(n)` can be referred from outside of the `for` loop:

```py
    while i < max_iter and error > tol:

        for i in range(n):
            v_next[i] = np.max(mcm.state_action_values(i, v))

        error = np.max(np.abs(v_next - v))
        i += 1  # <- This is always equal to n

        v[:] = v_next
```

Try this:

```py
for i in range(5):
    pass

print(i)
```

You will get `4`.

The index variable for the `for` loop (or that for the `while` loop) has to be something other than `i`.

Smit-create:
Fixes #175 
Fixes #176 

cc @jstac

### DIFF
--- a/lectures/_static/lecture_specific/coleman_policy_iter/solve_time_iter.py
+++ b/lectures/_static/lecture_specific/coleman_policy_iter/solve_time_iter.py
@@ -19,8 +19,7 @@ def solve_model_time_iter(model,    # Class with model information
 
     if error > tol:
         print("Failed to converge!")
-
-    if verbose:
+    elif verbose:
         print(f"\nConverged in {i} iterations.")
 
     return σ_new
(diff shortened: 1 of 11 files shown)


jstac:
Many thanks @Smit-create .

I think `if verbose` should be `elif verbose` --- we only want this print to be triggered if `error <= tol`  (see #175)

Smit-create:
> I think `if verbose` should be `elif verbose`

Yeah, you are right. Thanks for pointing out!

jstac:
Thanks @Smit-create , nice work.

PR #285, opened 2022-12-06
ENH: minor edits of new lecture mix-model

mmcky:
- [x] adjust equation references from using `tag` to equation links


mmcky:
@thomassargent30 this PR makes some minor adjustments to equation numbering to `eq` references rather than using `latex tags`

PR #291, opened 2022-12-10
MAINT: tidy up two_actions

mmcky:


mmcky:
Issue with this lecture is resolved in #293

PR #296, opened 2022-12-18
FIX: Adjust for quantecon==0.6.0

mmcky:
Migrates #286 to be local (cc: @Smit-create)

### DIFF
--- a/lectures/markov_asset.md
+++ b/lectures/markov_asset.md
@@ -313,6 +313,7 @@ The next figure shows a simulation, where
 * $g_t = \exp(X_t)$, so that $\ln g_t = X_t$ is the growth rate.
 
 ```{code-cell} ipython
+n = 7
 mc = qe.tauchen(n, 0.96, 0.25)
 sim_length = 80
 
@@ -962,7 +963,7 @@ $$
 Consider the following primitives
 
 ```{code-cell} python3
-n = 5
+n = 5  # Size of State Space
 P = np.full((n, n), 0.0125)
 P[range(n), range(n)] += 1 - P.sum(1)
 # State values of the Markov chain
--- a/lectures/markov_asset.md
+++ b/lectures/markov_asset.md
@@ -53,7 +53,7 @@ An asset is a claim on one or more future payoffs.
 
 The spot price of an asset depends primarily on
 
-* the anticipated  income stream  
+* the anticipated  income stream
 * attitudes about risk
 * rates of time preference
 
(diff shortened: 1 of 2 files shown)


mmcky:
@Smit-create 

I am getting this warning

```
/tmp/ipykernel_8015/2878916857.py:2: UserWarning: The API of tauchen has changed from `tauchen(rho, sigma_u, b=0., m=3, n=7)` to `tauchen(n, rho, sigma, mu=0., n_std=3)`. To find more details please visit: https://github.com/QuantEcon/QuantEcon.py/issues/663.
  mc = qe.tauchen(n, 0.96, 0.25)
```

in the preview of `markov_asset`

Do we need to run through an update the imports so these warnings are not presented?

mmcky:
I have tagged this as `high-priority` as `quantecon==0.6.0` is now available on `PyPI`

Smit-create:
> 

These warnings are expected. This warns users about this breaking change. It would be good to have atleast for sometime.

mmcky:
> These warnings are expected. This warns users about this breaking change. It would be good to have atleast for sometime.

I was expecting them to show up when the old format was used (rather than the new format). I think if a user is using what the new `api` will be there shouldn't be a warning issued right?

mmcky:
> > These warnings are expected. This warns users about this breaking change. It would be good to have atleast for sometime.
> 
> I was expecting them to show up when the old format was used (rather than the new format). I think if a user is using what the new `api` will be there shouldn't be a warning issued right?

@Smit-create is this warning showing up due to the import structure. If we update the imports to the new location will that remove this warning? Or will it show up whenever this function is used?

Smit-create:
> is this warning showing up due to the import structure.

No, this is because of changes in `tauchen`.

> Or will it show up whenever this function is used?

Yes, that's right. Can we keep these warnings for one version and remove them in the upcoming versions? This reason is that there is not much clear difference between the old and new APIs such that we can throw warnings only to users of the old API.

Smit-create:
Or else if we strictly want this warning to show only when the user uses the old API, one minor difference which I noticed is that:
Previously, the first argument of `tauchen` was float and now it should be an integer. We can throw this warning whenever we encounter a float at the first argument but this is still a bit hackish.

mmcky:
> Or else if we strictly want this warning to show only when the user uses the old API, one minor difference which I noticed is that:
> Previously, the first argument of `tauchen` was float and now it should be an integer. We can throw this warning whenever we encounter a float at the first argument but this is still a bit hackish.

Roger that - changes in positional arguments are hard to track at the function level and we haven't changed the name of the function so hard to identify. We could use the function signature (based on types) to identify old usage patterns though. I don't think that's a bad idea as it is confusing to get a warning when you have updated the order.

mmcky:
thanks for the updates @Smit-create I am going to merge this PR now so the cache will run using the latest quantecon release.

Issue #177, opened 2021-07-26
Open a file with `with`?


oyamad:
A discussion from my undergrad seminar:

[Tutorial](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files) says:

> It is good practice to use the with keyword when dealing with file objects.

Should this be followed in
* https://github.com/QuantEcon/lecture-python.myst/blob/main/lectures/short_path.md#exercise-1-1 and
* https://github.com/QuantEcon/lecture-python.myst/blob/main/lectures/finite_markov.md#exercise-2-1 ?

mmcky:
thanks @oyamad this is good feedback. 

Context managers are good in practice as they ensure `open` files are `closed` etc.

https://book.pythontips.com/en/latest/context_managers.html#context-managers

We do have `infile.close()` in that example as the file is built line-by-line. 
But if the `python` tutorial says we should do -- maybe we should :-).

PR #302, opened 2022-12-28
Update context manager

HumphreyYang:
This PR addresses #177 by adding a context manager `with` to the two aforementioned lectures.


mmcky:
thanks @HumphreyYang.

Issue #300, opened 2022-12-28
[wealth_dynamics] Rewrite Main Exercise Using Google JAX


Smit-create:
Writing JAX code for: https://python.quantecon.org/wealth_dynamics.html

#298

PR #304, opened 2022-12-30
JAX: Add JAX implementation for wealth dynamics

Smit-create:
#300

### DIFF
--- a/lectures/wealth_dynamics.md
+++ b/lectures/wealth_dynamics.md
@@ -323,7 +323,7 @@ def update_states_jax(arrays, wdy, size, rand_key):
 update_states_jax = jax.jit(update_states_jax, static_argnums=(2,))
 ```
 
-Here’s function to simulate the time series of wealth for in individual households.
+Here’s function to simulate the time series of wealth for individual households.
 
 ```{code-cell} ipython3
 def wealth_time_series_jax(w_0, n, wdy, size, rand_seed=1):
@@ -389,6 +389,30 @@ w_jax_result = wealth_time_series_jax(wdy.y_mean, ts_length, wdy, size)
 fig, ax = plt.subplots()
 ax.plot(w_jax_result)
 plt.show()
+```
+
+Now here’s function to simulate a cross section of households forward in time.
+
+```{code-cell} ipython3
+def update_cross_section_jax(w_distribution, shift_length, wdy, size, rand_seed=2):
+    """
+    Shifts a cross-section of household forward in time
+
+    * wdy: NamedTuple Model
+    * w_distribution: array_like, represents current cross-section
+
+    Takes a current distribution of wealth values as w_distribution
+    and updates each w_t in w_distribution to w_{t+j}, where
+    j = shift_length.
+
+    Returns the new distribution.
+    """
+    new_dist = wealth_time_series_jax(w_distribution, shift_length, wdy, size, rand_seed)
+    new_distribution = new_dist[-1, :]
+    return new_distribution
+
+# Create the jit function
+update_cross_section_jax = jax.jit(update_cross_section_jax, static_argnums=(1,3,))
 ```
 ## Implementation using Numba
 
@@ -558,7 +582,26 @@ wdy = WealthDynamics()
 
 ts_length = 200
 w = wealth_time_series(wdy, wdy.y_mean, ts_length)
+```
 
+```{code-cell} ipython3
+%%time
+
+w = wealth_time_series(wdy, wdy.y_mean, ts_length)
+```
+
+```{code-cell} ipython3
+%%time
+
+# Check the time for 2nd execution
+w = wealth_time_series(wdy, wdy.y_mean, ts_length)
+```
+
+Notice the time difference between the `wealth_time_series` and `wealth_time_series_jax`
+
+
+
+```{code-cell} ipython3
 fig, ax = plt.subplots()
 ax.plot(w)
 plt.show()
@@ -575,7 +618,24 @@ Let's look at how inequality varies with returns on financial assets.
 The next function generates a cross section and then computes the Lorenz
 curve and Gini coefficient.
 
+Let's first write the function that uses the jax implementation and for the numba.
+
 ```{code-cell} ipython3
+# Uses jax
+def generate_lorenz_and_gini_jax(wdy, num_households=100_000, T=500):
+    """
+    Generate the Lorenz curve data and gini coefficient corresponding to a
+    WealthDynamics mode by simulating num_households forward to time T.
+    """
+    size = (num_households, )
+    ψ_0 = jnp.full(size, wdy.y_mean)
+    ψ_star = update_cross_section_jax(ψ_0, T, wdy, size)
+    ψ_star = jax.device_get(ψ_star)  # get back to numpy form
+    return qe.gini_coefficient(ψ_star), qe.lorenz_curve(ψ_star)
+```
+
+```{code-cell} ipython3
+# Uses numba
 def generate_lorenz_and_gini(wdy, num_households=100_000, T=500):
     """
     Generate the Lorenz curve data and gini coefficient corresponding to a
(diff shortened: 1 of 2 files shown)


Smit-create:
Hi @HumphreyYang ,

This seems to be ready for the first review. Please have a look. Thanks

lectures/wealth_dynamics.md
```diff
@@ -250,7 +262,168 @@ their wealth.
 We are using something akin to a fixed savings rate model, while
 acknowledging that low wealth households tend to save very little.
 
-## Implementation
+## Implementation using JAX
```
HumphreyYang, line 265: Thanks for the update @Smit-create. The code looks great to me. One improvement we could potentially make is to hide the code using `numba` using 

```
---
tags: [hide-input]
---
```

and only compare the speed when making the final comparison. 

I think it is clear to readers when comparing the performance of two implementations in my case since there is only one function involved, but in this case, having two implementations throughout the lecture may confuse readers and obscure the focus of the lecture. Please me know what you think :) (also CC @jstac)

HumphreyYang (commented):
Thanks for the update @Smit-create. The code looks great to me. One improvement we could potentially make is to hide the code using `numba` using 

```
---
tags: [hide-input]
---
```

and only compare the speed when making the final comparison. 

I think it is clear to readers when comparing the performance of two implementations in my case since there is only one function involved, but in this case, having two implementations throughout the lecture may confuse readers and obscure the focus of the lecture. Please let me know what you think :) (also CC @jstac)

lectures/wealth_dynamics.md
```diff
@@ -83,6 +83,15 @@ import jax.numpy as jnp
 from collections import namedtuple
 ```
 
+Let's check the backend used by jax and the devices available
+
+```{code-cell} ipython3
+# Check if JAX is using GPU
+print("jax backend: {}".format(jax.lib.xla_bridge.get_backend().platform))
+# Check the devices available for JAX
+print(jax.devices())
```
HumphreyYang: I think we can use 

```
print(f"jax backend: {jax.devices()[0].platform}")
```

in our PRs for simplicity and consistency.

jstac: Thanks @HumphreyYang , simple is good.

@Smit-create , please change `jax` to `JAX` in the text.

Smit-create: Thanks, I will change that accordingly.

Smit-create:
> Please let me know what you think :)

This sounds like a nice idea, let's try that out and see how it looks.

jstac:
@Smit-create and @HumphreyYang, please let me know when this is ready for another review.

HumphreyYang:
> > Please let me know what you think :)
> 
> This sounds like a nice idea, let's try that out and see how it looks.

@Smit-create I think it looks cleaner and simpler with Numba code hidden. How do you feel about this change?

lectures/wealth_dynamics.md
```diff
@@ -250,11 +262,242 @@ their wealth.
 We are using something akin to a fixed savings rate model, while
 acknowledging that low wealth households tend to save very little.
 
-## Implementation
+## Implementation using JAX
+
+Let's define a Model to represent the wealth dynamics.
```
jstac: Please change "Model" to "model" and the same below.

lectures/wealth_dynamics.md
```diff
@@ -250,11 +262,242 @@ their wealth.
...
+
+    R = wdy.c_r * jnp.exp(zp) + jnp.exp(wdy.μ_r + wdy.σ_r * jax.random.normal(subkey[1], shape=size))
+    wp += (w >= wdy.w_hat) * R * wdy.s_0 * w
+    return wp, zp
+
+# Create the jit function
+update_states_jax = jax.jit(update_states_jax, static_argnums=(2,))
+```
+
+Here’s function to simulate the time series of wealth for individual households using `for` loop and JAX.
```
jstac: using *a* `for` loop

lectures/wealth_dynamics.md
```diff
@@ -250,11 +262,242 @@ their wealth.
...
+```
+
+Running the above function again will be even faster because of JAX's JIT.
+
+```{code-cell} ipython3
+%%time
+
+# 2nd time is expected to be very fast because of JIT
+w_jax_result = wealth_time_series_for_loop_jax(wdy.y_mean, ts_length, wdy, size)
+```
```
jstac, line 406: does this need `block_until_ready`?  

I guess not but just checking.

jstac:
Looking at the current deployment https://63b52771da28e92bd5e7733b--nostalgic-wright-5fa355.netlify.app/wealth_dynamics.html shows that one numba code cell is not hidden, while the others are.  Is this intensional?

jstac:
Should we compare the timing of `update_cross_section_jax` and `update_cross_section`?

I suspect that, on a large cross-section, the JAX code will be much faster...

jstac:
Does this sentence need to be adjusted/removed?

"Notice how effectively Numba has implemented multithreading for this routine: all 8 CPUs on our workstation are running at maximum capacity (even though four of them are virtual)."

I think we are using JAX at this point.

jstac:
Many thanks @Smit-create .  Please see my comments above.

@mmcky , if you have time please have a quick glance at this once @Smit-create has pushed changes.

lectures/wealth_dynamics.md
Smit-create: Let's add that. The documentation for benchmarking also recommends using it: https://jax.readthedocs.io/en/latest/faq.html#benchmarking-jax-code

Smit-create:
> shows that one numba code cell is not hidden, while the others are. Is this intentional?

I thought to keep it unhidden so that the below cell that measures time can relate to it. But let's hide all the numba code and just keep the JAX available readily.

> Should we compare the timing of update_cross_section_jax and update_cross_section?

It's already been compared using `generate_lorenz_and_gini_jax` (while plotting) and shows about a ~30s difference.

> "Notice how effectively Numba has implemented multithreading for this routine: all 8 CPUs on our workstation are running at maximum capacity (even though four of them are virtual)."

Let's remove this completely along with the image?

mmcky:
thanks @Smit-create, @HumphreyYang and @jstac 

Is the purpose of hiding the `Numba` code to clean up the lecture and enable those interested to see the code. It is not the approach often taken relative to our other lectures which tends to always show the code but I see the application here. 

This is what it looks like in the preview

Screenshot 2023-01-13 at 1 05 33 pm

mmcky:
For the text in the lecture

```
Notice the time difference between the wealth_time_series and wealth_time_series_jax
```

Could we use the `glu` feature to have the values here? It is a statement that requires the reader to find the values and compare them.

jstac:
Yes, I think it looks nice to have the numba code hidden, and I can understand the motivation, but at the same time it might be confusing for the reader --- why is some code hidden and some not?

Also, if we hide numba code here, then we should do it elsewhere when we use JAX --- but then it's a burden and possibly unclear exactly what we should hide.

Having thought all this through, I suggest that we show the numba code rather than hiding it and do the same for the remaining lectures.

Smit-create:
> I suggest that we show the numba code rather than hiding it and do the same for the remaining lectures.

Sure, will do that.

> Could we use the `glue` feature to have the values here?

Yes, added it, thanks!

jstac:
Thanks @Smit-create 

@HumphreyYang just checking that you saw the discussion above regarding hiding / not hiding.

Could you please give this PR one more careful look over and then let me know when you are done.  After that I think we'll be ready to merge.

HumphreyYang:
> Thanks @Smit-create 
> 
> 
> 
> @HumphreyYang just checking that you saw the discussion above regarding hiding / not hiding.
> 
> 
> 
> Could you please give this PR one more careful look over and then let me know when you are done.  After that I think we'll be ready to merge.

Hi @jstac,

Before that, could you please have a look at the [GPU Warning and JAX imports](https://63bd6252e877bd27ca902c25--nostalgic-wright-5fa355.netlify.app/kesten_processes.html) I pushed in my PR so that @Smit-create and I can have a consistent GPU warning?

jstac:
Thanks @HumphreyYang , good idea.  Here's what I suggest for the GPU warning, which slightly modifies the version of @HumphreyYang :

This lecture is built using hardware that has access to a GPU and uses JAX for GPU programming. As a result, the lecture will be slower when running on a machine without a GPU.

Free GPUs are available on Google Colab.  To use this option, please click on the play icon top right, select Colab, and set the runtime environment to include a GPU.

Alternatively, if if you have your own GPU, you can follow the [instructions](https://github.com/google/jax#installation) for for installing JAX with GPU support.

Smit-create:
The PR is ready now. The failure is because of this [diff](https://github.com/QuantEcon/lecture-python.myst/pull/301/files#diff-35b74923e7e9fd8e1da5e4a8f07ad1b655c021d356ccf71902437ce1fde25c9b). Once we merge that PR, I'll rebase this on top of that and that should fix the CI.

jstac:
Thanks @Smit-create 

- Please see the requested changes to GPU Warning in #301 --- sorry for repeated changes and thanks for your patience
- Is there a working netlify deployment I can look at here?

Smit-create:
> * Please see the requested changes to GPU Warning

Thanks for the review @jstac. I have updated the diff.

> Is there a working netlify deployment I can look at here?

The last commit (10th commit) was failing on CI and so I suppose it won't be deployed. I have pushed a new commit and let's see if that works.

Smit-create:
The tests pass and the deployed version is available at https://63c622d321fc755cc8652567--nostalgic-wright-5fa355.netlify.app/

jstac:
Great work @Smit-create , many thanks.  Merging.

PR #307, opened 2023-01-16
Fix small typos in McCall Q learning

jstac:


jstac:
There are some hard-coded equation numbers that need to be fixed --- for example, 

- "described in equation system (4) lets"
- "with our Q-learning recursion (3)"
- "Q-table defined in equation (2) automatically "

jstac:
Hi @mmcky , what do you think is happening here (link checker fail)?

HumphreyYang:
Hi @jstac, 

I saw my link checker passed, so I rerun the link checker in your PR. I think the server of that website was overloaded, but it is now passing.

mmcky:
@mmcky sometimes the link checker fails if a server takes too long to respond. It's annoying part hard to fix. Most of the time it returns a reliable result. FWIW the link checker only runs on the first PR creation so if there are any edits it won't block the PR.

mmcky:
thanks @shlff did you address all 3 equation reference updates?

jstac:
Many thanks @shlff .  Merging.

PR #325, opened 2023-03-23
Fix small errors and add one tag

HengchengZhang:
This PR 

- Fixes the wrong visualization of matrix b and c
https://github.com/QuantEcon/lecture-python.myst/blob/604d707e132261bc9071bc5f9faf9e1c4e35f670/lectures/newton_method.md?plain=1#L545-L570

- Adds a tag before 
https://github.com/QuantEcon/lecture-python.myst/blob/604d707e132261bc9071bc5f9faf9e1c4e35f670/lectures/newton_method.md?plain=1#L466

  to link the new Jax version of Newton's Method back to the same topic.

Also, @jstac I'm confused with the interpretation of the excess demand function in python.
In line 482, exp( - A @ p) is used to represent the $e^{-a_{i0} p_0} + e^{-a_{i1} p_1}, i = 0, 1$ part in the excess demand function but I think they are not mathematically equivalent.

### DIFF
--- a/lectures/newton_method.md
+++ b/lectures/newton_method.md
@@ -470,7 +470,7 @@ Let's start by computing the market equilibrium of a two-good problem.
 We consider a market for two related products, good 0 and good 1, with
 price vector $p = (p_0, p_1)$
 
-Supply of good $i$ at price $p$,
+Supply of good $i$ at price $p$ is,
 
 $$ 
 q^s_i (p) = b_i \sqrt{p_i} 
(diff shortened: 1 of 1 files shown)


jstac:
@mmcky and @HengchengZhang , could someone see what the error is here please?  Once it's building I'll take a look at the preview.

mmcky:
@HengchengZhang I will check you are a collaborator so you can use a local branch. The preview generation etc. won't work from a `fork` due to permission issues and accessing credentials.

mmcky:
@HengchengZhang the collaborator page isn't loaded right this minute. I will loop around in 30min or so.

HengchengZhang:
Thanks @mmcky, I will try it again later.

PR #327, opened 2023-04-01
ENH: Upgrade docker with anaconda==2023.03

mmcky:
- [x] check `jax` and `gpu` computing
- [x] check compatibility with `anaconda==2023.03`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,13 +40,13 @@ jobs:
       - name: Display Pip Versions
         shell: bash -l {0}
         run: pip list
-      - name: Download "build" folder (cache)
-        uses: dawidd6/action-download-artifact@v2
-        with:
-          workflow: cache.yml
-          branch: main
-          name: build-cache
-          path: _build
+      # - name: Download "build" folder (cache)
+      #   uses: dawidd6/action-download-artifact@v2
+      #   with:
+      #     workflow: cache.yml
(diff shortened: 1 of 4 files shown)


mmcky:
- updated `3d` projections to

```
ax = plt.figure(figsize = (10, 6)).add_subplot(projection='3d')
```

mmcky:
@jstac @HumphreyYang this is now showing as running on the `gpu` after installing `jax == 0.4.8` in the docker container and connecting it to the latest `cuda 12.1` release

mmcky:
@HumphreyYang @Smit-create if you have any time can you please take a look through the preview deployment above and let me know if you see any issues. This PR is moving to:

1. `anaconda==2023.03` which is the latest set of packages
2. `jax[cuda12_pip]`
3. `cuda 12.1` docker container from Nvidia

I noticed that the drivers being reported by `nvidia-smi` are `470.161.03` which appears to be an issue for `cuda 12` so I don't quite understand but the lectures are reporting `gpu` backend. 

I am doing some research into how to update the drivers on `aws` instances `p3.2xlarge`

Smit-create:
Thanks @mmcky, The execution of Wealth Dynamics looks fine to me. The results are errorless and as expected, JAX is faster than numba.

mmcky:
thanks @Smit-create and @HumphreyYang

PR #330, opened 2023-04-02
Remove JAX content in Kesten Processes

### LINKED ISSUE #329: [kesten processes] large memory usage
It looks like the `kesten_processes` lecture is requesting more than 10Gb of memory when using `jax` in the examples. 

@HumphreyYang the example looks like it preallocates some very large matrices. Perhaps we can revise this example to be more memory efficient by generated random variables within the loop rather than predefined matrices of dimension `1_000_000`

```
/home/mmcky/work/quantecon/lecture-python.myst/lectures/kesten_processes.md: Executing notebook using local CWD [mystnb]                         
2023-04-02 13:26:43.851719: W external/xla/xla/service/hlo_rematerialization.cc:2194] Can't reduce memory use below 7.33GiB (7871741952 bytes) by rematerialization; only reduced to 9.33GiB (10019999952 bytes)
2023-04-02 13:26:59.843560: W external/tsl/tsl/framework/bfc_allocator.cc:485] Allocator (GPU_0_bfc) ran out of memory trying to allocate 9.31GiB (rounded to 10000032512)requested by op 
2023-04-02 13:26:59.843781: W external/tsl/tsl/framework/bfc_allocator.cc:497] *___________________________________________________________________________________________________
2023-04-02 13:26:59.846908: E external/xla/xla/pjrt/pjrt_stream_executor_client.cc:2432] Execution of replica 0 failed: RESOURCE_EXHAUSTED: Out of memory while trying to allocate 10000032512 bytes.
BufferAssignment OOM Debugging.
BufferAssignment stats:
             parameter allocation:         0B
              constant allocation:         8B
        maybe_live_out allocation:    3.81MiB
     preallocated temp allocation:    9.31GiB
  preallocated temp fragmentation:         0B (0.00%)
                 total allocation:    9.32GiB
              total fragmentation:    3.84MiB (0.04%)
```

HumphreyYang:
This PR resolves #329 by reducing the sample size in the Kesten Process exercise.


jstac:
Hi @HumphreyYang ,

For the for loop at least, I think we want to generate a fresh array of shocks at each point in time, similar to the JAX simulation in https://github.com/QuantEcon/rse_comp_econ_2023/blob/main/notebooks/european_option.ipynb

Then memory usage won't blow up with changes to maturity or the time scale.

This might be more problematic with scan --- if so, please mention the issue and keep the sample size for scan small.

HumphreyYang:
Hi @jstac, 

Many thanks for the suggestions.

My idea is to shrink the size of all of them if we need to reduce the size so readers can have a good idea when we run the comparison, but before that, I will give it a try to fit the original exercise into the memory.

HumphreyYang:
Hi @mmcky,

I did not replicate the memory issue you have potted in #329. Could you give me some more context about how this issue occurred? Did the issue happen on EC2?

Many thanks in advance.

mmcky:
> Hi @mmcky,
> 
> I did not replicate the memory issue you have potted in #329. Could you give me some more context about how this issue occurred? Did the issue happen on EC2?
> 
> Many thanks in advance.

@HumphreyYang this happened on the `quantecon` gpu server and I used these lectures as a test case. It only has 10Gb (so only 7.8 is available to programs and this lecture is requesting 9.7) of memory. The `V100` on aws has 16Gb of RAM so probably not an issue there. 

I think @jstac suggestion is a good one and is a lot more memory efficient. It would be interesting to know if moving the draws into the inner loop would cause a penalty hit (over pre-allocation).

jstac:
Yes, please follow my proposed approach.

jstac:
See my note above on why memory will be stable with respect to the time horizon

HumphreyYang:
> > Hi @mmcky,
> > I did not replicate the memory issue you have potted in #329. Could you give me some more context about how this issue occurred? Did the issue happen on EC2?
> > Many thanks in advance.
> 
> @HumphreyYang this happened on the `quantecon` gpu server and I used these lectures as a test case. It only has 10Gb (so only 7.8 is available to programs and this lecture is requesting 9.7) of memory. The `V100` on aws has 16Gb of RAM so probably not an issue there.
> 
> I think @jstac suggestion is a good one and is a lot more memory efficient. It would be interesting to know if moving the draws into the inner loop would cause a penalty hit (over pre-allocation).

Many thanks @mmcky and @jstac. 

I have already put the iterations into the `for` loop. Also, there is a way to turn off preallocation in JAX or make it on-demand, and I would like to try it on `lax.scan`, which is the one that mostly likely to cause the memory issue. I think I may need to do some memory profiling on the device itself to see how much adjustment I need to have on `lax.scan` version. Please kindly let me know if it is possible.

jstac:
Thanks @HumphreyYang .  I suggest that you show the scan one with a warning that memory will blow up with large horizons.  If you still want to do a horse race with the for loop, then please run both with small n (short time horizon).

HumphreyYang:
Many thanks for the suggestion @jstac.

Before committing my change, one thing I noticed is that there is a significant performance loss (around one minute) in compilation time when more computations are moved into the `for` loop in the jitted function. Please kindly see my note [here](https://gist.github.com/HumphreyYang/5dadcbabfaeb2747e7ce2ed8db456732). This motivated me to put the sampling process outside the `for` loop in my previous version.

JAX also included this in the [documentation](https://jax.readthedocs.io/en/latest/faq.html#jit-decorated-function-is-very-slow-to-compile) and suggested that once the computation in the `for` loop is too large, we should switch to `lax.scan`.  

Please kindly let me know what is your experience on this.

HumphreyYang:
Hi @jstac,

I have cut the runtime for the `for` loop version to 4 secs compared to 1.5 mins before. 

Please kindly review the change, and I will proceed to cut the time horizon if it looks good to you.

Many thanks in advance.

jstac:
Nice work @HumphreyYang ,  very good.

I think this is good to go.  Please remove the do not merge label and ping me when you are happy.

mmcky:
@HumphreyYang I just ran this on the smaller `gpu` server in my office and still getting memory issues

```
/home/mmcky/work/quantecon/lecture-python.myst/lectures/kesten_processes.md: Executing notebook using local CWD [mystnb]   
2023-04-06 11:26:57.756605: W external/xla/xla/service/hlo_rematerialization.cc:2194] Can't reduce memory use below 7.33GiB (7871741952 bytes) by rematerialization; only reduced to 9.32GiB (10004000016 bytes)
2023-04-06 11:27:07.939062: W external/tsl/tsl/framework/bfc_allocator.cc:485] Allocator (GPU_0_bfc) ran out of memory trying to allocate 11.18GiB (rounded to 12000005376)requested by op 
2023-04-06 11:27:07.939291: W external/tsl/tsl/framework/bfc_allocator.cc:497] *___________________________________________________________________________________________________
2023-04-06 11:27:07.939674: E external/xla/xla/pjrt/pjrt_stream_executor_client.cc:2432] Execution of replica 0 failed: RESOURCE_EXHAUSTED: Out of memory while trying to allocate 12000005200 bytes.
BufferAssignment OOM Debugging.
BufferAssignment stats:
             parameter allocation:         0B
              constant allocation:        24B
        maybe_live_out allocation:    3.81MiB
     preallocated temp allocation:   11.18GiB
  preallocated temp fragmentation:        48B (0.00%)
                 total allocation:   11.18GiB
              total fragmentation:    3.82MiB (0.03%)
```

I will try to arrange access to the server for you over the weekend to do some testing but please liaise with @jstac re: when you plan to use it so we don't get `gpu` request collisions :-)

jstac:
I propose that, with the `scan` implementation,

1. We only use a small sample size --- maybe 1/10th of the for loop version.
2. We warn users that memory usage will blow up quickly --- and this is why we are using such a small sample relative to the for loop
3. If we want to do a horse race with the for loop, then the for loop function should be run a second time, separately, after `scan`, with the same small sample as `scan`, purely as a way to compare the for loop and `scan`.

Note, however, that this is a lecture on heavy tails, so small samples will not be enough to learn about the underlying distribution.  It could be a good idea to have a warning message about that.

HumphreyYang:
Hi @jstac, 

This is ready to merge. The JAX code has been moved to the JAX lecture in PR [here](https://github.com/QuantEcon/lecture-jax/pull/29#issuecomment-1536952130)

jstac:
Perfect.  Many thanks @HumphreyYang .

PR #333, opened 2023-04-12
ENH: Use common file for GPU admonition

mmcky:
This PR moves the GPU admonition to `_admonitions/gpu.md`

- [x] wealth_dyanmics
- [x] kesten_processes

@jstac do you agree with this approach. 

The alternative would be to add a more subtle tag such as in https://github.com/QuantEcon/lecture-python-programming.myst/pull/255#issuecomment-1499737267

fixes https://github.com/QuantEcon/meta/issues/75


mmcky:
@jstac if you think this is a better setup I can merge and update the QE manual.

jstac:
@mmcky Thanks.  These lectures are having JAX stripped out so I think it's correct to remove the tag.  (I'm not sure why we would want a more subtle one if the JAX code is being removed?  Sorry if I've misunderstood.)

mmcky:
@jstac this is essentially a proof of concept to setup `quantecon` standard `admonitions` such as the GPU message for files that make use of the `gpu`. Indeed -- this is getting migrated to `lecture-jax` but this is a style question for you -- so we only have one place to update the admonition text that is shared across files. 

If you like this style I will update our `style guide` with default locations for `_admonitions` in lectures.

jstac:
Yes, good plan @mmcky, thanks!

Note that @Smit-create has already removed JAX from one of these lectures though --- see #335 

It might be gone from the other one too --- I can't remember.

mmcky:
Roger that. Thanks @jstac

PR #335, opened 2023-04-18
Remove JAX from Wealth dynamics

Smit-create:
Reverts #304 
https://github.com/QuantEcon/lecture-jax/pull/22

### DIFF
--- a/lectures/wealth_dynamics.md
+++ b/lectures/wealth_dynamics.md
@@ -24,8 +24,8 @@ kernelspec:
 ```
 
 ```{seealso}
-**GPU:** A version of this lecture which makes use of [jax](https://jax.readthedocs.io) to run the code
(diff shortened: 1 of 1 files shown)


jstac:
Great, thanks @Smit-create , much appreciated.

@mmcky Would you mind to handle this one?

mmcky:
Thanks @Smit-create this looks create. Now that jax.quantecon.org is available we can go ahead and merge this. 

- [x] I added a `seealso` admonition to provide a link to the `GPU` enabled version. 

Screenshot 2023-05-09 at 12 41 31 pm

@jstac I think we could improve on this -- but is this OK for now?

jstac:
That looks great to me.   I would be happy with One small thing is that "jax" should be capitalized.

PR #349, opened 2023-06-07
Remove `heavy_tails` and Add `intersphinx_mapping` to the Intro Series

### LINKED ISSUE #336: Remove heavy tailed distributions lecture from this site
The lecture `heavy_tails.md` is being moved to the undergraduate series and needs to be cut from this lecture series when the undergraduate series goes live.

HumphreyYang:
This resolves #336.


jstac:
Thanks @HumphreyYang .

@mmcky could you please handle this?

mmcky:
@HumphreyYang could you please add in a `heavy_tails` page that is similar to the `jax` page above. We can then open an issue to completely remove the page in 3 months or so. I think this is a better way to go -- so if anyone is linked to it they have time to update.

mmcky:
thanks @HumphreyYang

Issue #169, opened 2021-07-18
[lake_model] Comments


oyamad:
* This part https://github.com/QuantEcon/lecture-python.myst/blame/main/lectures/lake_model.md#L260-L274 is not very wise (computing the stationary distribution of a 2-state (column-)stochastic matrix by iteration). As shown in the [Finite Markov Chains](https://python.quantecon.org/finite_markov.html#example) chapter, and to be discussed in the [current chapter](https://python.quantecon.org/lake_model.html#dynamics-of-an-individual-worker), it can be simply computed exactly (up to floating points errors) by
  ```py
  def rate_steady_state(self):
      x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
      return x / x.sum()
  ```

* The discussion in "[Aggregate Dynamics](https://python.quantecon.org/lake_model.html#aggregate-dynamics)":
  This part is hard to read. The same discussion as in "Finite Markov Chains" is given in a different language without any indication: From the discussion in "Finite Markov Chains", we know that
  * the (column-)stochastic matrix `A_hat` has a stationary distribution (or equivalently, it has a nonnegative eigenvector with eigenvalue one); and
  * `A_hat` being (irreducible and) aperiodic (or equivalently, the other eigenvalues are less than one in magnitude), from any initial distribution we have convergence to the (unique) stationary distribution.
  
  In my view, this new language (with eigenvalues) is not necessary, and it would be enough to refer to the previous discussion in "Finite Markov Chains" (as to be done [below](https://python.quantecon.org/lake_model.html#ergodicity)).

* There are a few places where the inner product of two vectors (1d-ndarrays) `a` and `b` is computed by
  ```py
  np.sum(a * b)
  ```

  instead of 
  ```py
  a @ b
  ```

  Is there any purpose for this?

jstac:
Thanks @oyamad .  I agree.  I'll fix this when I get some time.

shlff:
Thanks @jstac and @oyamad . 

My comment on point 1 is that It is a smart change that makes the most of the analytical solution of the stationary distribution of the Markov chain when the stochastic matrix is positive. 

For example, let the stochastic matrix be

$$
P = \left(
        \begin{matrix}
            1 - \lambda & \lambda \\
            \alpha & 1 - \alpha
        \end{matrix}
    \right)
$$

if $\alpha \in (0, 1)$ and $\lambda \in (0, 1)$, then $P$ has a unique stationary distribution.

[...]

PR #355, opened 2023-07-10
[LakeModel] Simplify computation of the steady state

mnshkw:
Compute the steady state by the straightforward formula for a 2x2 Markov chain instead of iteration.
See also: #169.

### DIFF
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-2. provide a method to compute the steady state vector of employment and unemployment rates $\bar x$ using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
+2. provide a method to compute the steady state vector $\bar x$ of employment and unemployment rates using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
 
 Please be careful because the implied objects $g, A, \hat A$ will not change
 if you only change the primitives.
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-1. provide a method to compute the steady state of the rate (We will discuss it in section {ref}`dynamics_workers`)
+1. provide a method to compute the steady state of the rate (which we discuss in section {ref}`dynamics_workers`)
 
 Please be careful because the implied objects $g, A, \hat A$ will not change
 if you only change the primitives.
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -265,7 +265,7 @@ class LakeModel:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
+        x = np.array([self.A_hat[1, 0], self.A_hat[0, 1]])
         x /= x.sum()
         return x
 
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -1013,7 +1013,7 @@ class LakeModelModified:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
+        x = np.array([self.A_hat[1, 0], self.A_hat[0, 1]])
         x /= x.sum()
         return x
 
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -265,7 +265,7 @@ class LakeModel:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.array([self.A_hat[1, 0], self.A_hat[0, 1]])
+        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
         x /= x.sum()
         return x
 
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-2. provide a method to compute the steady state rate using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
+2. provide a method to compute the steady state vector of employment and unemployment rates $\bar x$ using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
 
 Please be careful because the implied objects $g, A, \hat A$ will not change
 if you only change the primitives.
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-1. provide a method to compute the steady state of the rate (discussed in {ref}`dynamics_workers`)
+1. provide a method to compute the steady state of the rate (We will discuss it in section {ref}`dynamics_workers`)
 
 Please be careful because the implied objects $g, A, \hat A$ will not change
 if you only change the primitives.
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -1013,7 +1013,7 @@ class LakeModelModified:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.array([self.A_hat[1, 0], self.A_hat[0, 1]])
+        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
         x /= x.sum()
         return x
 
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -265,12 +265,8 @@ class LakeModel:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.full(2, 0.5)
-        error = tol + 1
-        while error > tol:
-            new_x = self.A_hat @ x
-            error = np.max(np.abs(new_x - x))
-            x = new_x
+        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
+        x /= x.sum()
         return x
 
     def simulate_stock_path(self, X0, T):
--- a/lectures/lake_model.md
+++ b/lectures/lake_model.md
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-1. provide a method to compute the steady state of the rate
+1. provide a method to compute the steady state of the rate (discussed in {ref}`dynamics_workers`)
 
 Please be careful because the implied objects $g, A, \hat A$ will not change
 if you only change the primitives.
@@ -411,6 +411,7 @@ plt.tight_layout()
 plt.show()
 ```
 
+(dynamics_workers)=
 ## Dynamics of an Individual Worker
 
 An individual worker's employment dynamics are governed by a {doc}`finite state Markov process <finite_markov>`.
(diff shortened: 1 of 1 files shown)


jstac:
Thanks @mnshkw , appreciated.

@shlff , could you please do a first round review of this and also consider #169 ?

jstac:
Thanks @mnshkw , appreciated.

@shlff , could you please do a first round review of this and also consider #169 ?

The build error might need the attention of @mmcky

lectures/lake_model.md
```diff
@@ -265,12 +265,8 @@ class LakeModel:
...
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.full(2, 0.5)
-        error = tol + 1
-        while error > tol:
-            new_x = self.A_hat @ x
-            error = np.max(np.abs(new_x - x))
-            x = new_x
+        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
+        x /= x.sum()
```
shlff, line 269: Suggest
```
        x /= x.sum()
        return x
```
to 
```
        return x / x.sum()
```

shlff:
Thanks @mnshkw and @jstac . 

I quickly reviewed this change and issue #169 , which makes the most of the analytical solution of the stationary distribution of the Markov chain when the stochastic matrix is positive. 

For example, let the stochastic matrix be

$$
P = \left(
        \begin{matrix}
            1 - \lambda & \lambda \\
            \alpha & 1 - \alpha
        \end{matrix}
    \right)
$$

if $\alpha \in (0, 1)$ and $\lambda \in (0, 1)$, then $P$ has a unique stationary distribution.

It is a smart change, but the thing is that it cannot handle the case when $\alpha$ or $\lambda$ takes the boundary values, that is, $\alpha, \lambda$ takes $0, 1$.

I suggest that we turn this change into an exercise.

lectures/lake_model.md
oyamad: The former is in principle strictly better: `x / x.sum()` creates an intermediate array while `x /= x.sum()` is an in-place operation. (Of course, the difference is negligible for this size of the data.)

oyamad:
@shlff Good point. The proposed code does not work if `lambda = alpha = 0` (in your notation) (it does as long as at least one is strictly positive). But in this case, there is nothing to analyze?

shlff:
Thanks for clarifying it @oyamad . Sorry for my late reply.

I reckon there are two benefits of mentioning the extreme case. 

One is that we can treat the case as a limiting reference point for our analysis of varying values of ``lambda, alpha`` in lecture [lake_model](https://python.quantecon.org/lake_model.html#dynamics-of-an-individual-worker).

Additionally, mentioning this case enhances the audience's understanding of how we can apply the improved method, when it will fail, and how we can resolve the issue when it fails.

oyamad:
@shlff Thanks, but I am afraid I could not see your point. The arguments in that lecture all rely on the assumptions of irreducibility (`lambda >0`, `alpha > 0`) and aperiodicity (`(lambda, alpha) neq (1, 1)`).

The formula `psi^*[0] = alpha / (alpha + lambda)` fails if and only if `alpha = 0` and `lambda = 0`. In this case, any distribution is obviously a stationary distribution?

Could you elaborate?

shlff:
The following part of the lecture ``lake_model`` could be updated too:

https://github.com/QuantEcon/lecture-python.myst/blob/a4d4aaaf1fee7e09e39412a7c66c77f258b109c6/lectures/lake_model.md?plain=1#L1011-L1025

mmcky:
Execution issue is due to: https://github.com/orgs/QuantEcon/discussions/108

mmcky:
I need to fix #359 to fix the build

mmcky:
@shlff would you mind to have a final review of this PR? Is anything outstanding? 

Preview builds will not run as the source change is from an external fork.

shlff:
Thanks @mmcky . I will check this PR and get back to you by Friday 15th.

shlff:
Thanks @mmcky . After thorough thinking of the simplified approach and related lectures, whether we should adopt this change **here** depends on the goal of this lecture.

If we assume our audience of this lecture should have read the lecture [finite Markov chains](https://python.quantecon.org/finite_markov.html) and have a good grasp of basic concepts, such as ergodicity. In that case, this is a beautiful change.

Otherwise, I don't think it is easy for readers to follow the simplified computation with the information given above this change (that is, the context from sections [69.2.1](https://python.quantecon.org/lake_model.html#aggregate-variables) to [69.2.3](https://python.quantecon.org/lake_model.html#laws-of-motion-for-rates)) since we don't introduce anything about Markov chains and ergodicity there. 

I guess the writer wanted to make the key point that we can study dynamics without concepts like Markov and ergodicity, but the results could be very shaky. Therefore ergodicity could be a nifty concept because it makes our results solid.

Instead, I recommend introducing this simplification in an inheritance class to the original class after the section [Ergodicity](https://python.quantecon.org/lake_model.html#ergodicity). 

Since we already give the interpretation and mathematics of ergodicity and build the connection to [finite Markov chains](https://python.quantecon.org/finite_markov.html) in that section, it will be much easier and natural for the audience to come up with the simplified method next. 

No matter what goal we have, we should also apply this change to the corresponding method of a class in [one of the exercises of that lecture](https://python.quantecon.org/lake_model.html#exercises):

https://github.com/QuantEcon/lecture-python.myst/blob/a4d4aaaf1fee7e09e39412a7c66c77f258b109c6/lectures/lake_model.md?plain=1#L1011-L1025

jstac:
Thanks @shlff , please go ahead.

lectures/lake_model.md
```diff
@@ -265,12 +265,8 @@ class LakeModel:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.full(2, 0.5)
-        error = tol + 1
-        while error > tol:
-            new_x = self.A_hat @ x
-            error = np.max(np.abs(new_x - x))
-            x = new_x
+        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
```
oyamad, line 268: This is a typo...
```suggestion
        x = np.array([self.A_hat[1, 0], self.A_hat[0, 1]])
```

oyamad:
I am afraid I did not understand this argument:

> Additionally, I suggest that we add a note here to remind readers that we will discuss in the section [ergodicity](https://python.quantecon.org/lake_model.html#ergodicity) why we can calculate the steady state in this way and the conditions we should check to ensure the existence.

The code comes from the formula "inflow = outflow" at a steady state: `x[1] * P[1, 0] = x[0] * P[0, 1]`, with the sum being `1`. It has nothing to do with "ergodicity"; it is just the definition of a stationary distribution.

The reference should be to "[25.6. Stationary Distributions](https://python.quantecon.org/finite_markov.html#stationary-distributions)" instead.

jstac:
I agree with @oyamad .

shlff:
> The code comes from the formula "inflow = outflow" at a steady state: x[1] * P[1, 0] = x[0] * P[0, 1], with the sum being 1. It has nothing to do with "ergodicity"; it is just the definition of a stationary distribution.

> The reference should be to "[25.6. Stationary Distributions](https://python.quantecon.org/finite_markov.html#stationary-distributions)" instead.

Thanks @oyamad . I am sorry that's a typo. Actually, I refer it to [69.4. Dynamics of an Individual Worker](https://python.quantecon.org/lake_model.html#dynamics-of-an-individual-worker), where there is a discussion on stationary distribution.

lectures/lake_model.md
```diff
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-1. provide a method to compute the steady state of the rate
+1. provide a method to compute the steady state of the rate (We will discuss it in section {ref}`dynamics_workers`)
```
jstac: "We will discuss..." -> "which we discuss.."

jstac:
@mmcky , would you mind to fix the build?

lectures/lake_model.md
```diff
@@ -1016,12 +1013,8 @@ class LakeModelModified:
         --------
         xbar : steady state vector of employment and unemployment rates
         """
-        x = np.full(2, 0.5)
-        error = tol + 1
-        while error > tol:
-            new_x = self.A_hat @ x
-            error = np.max(np.abs(new_x - x))
-            x = new_x
+        x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])
```
oyamad, line 1016: Copied typo:
```suggestion
        x = np.array([self.A_hat[1, 0], self.A_hat[0, 1]])
```

mmcky:
@jstac this build cannot be fixed as it is originating from a fork (GitHub doesn't want to give out our EC2 credentials -- which is actually nice of them :-)) but it is a pain. I haven't been able to think of a good technical solution to this problem yet. It's only an issue for lectures that make use of `EC2` as the backend (i.e. GPU). 

I will make a replicate of the PR on a local branch so you can preview.

lectures/lake_model.md
mmcky:
```suggestion
1. provide a method to compute the steady state of the rate (which we discuss in section {ref}`dynamics_workers`)
```

lectures/lake_model.md
shlff: Thanks @oyamad . I am afraid that I cannot see why this is a typo.

The method with ``x = np.array([self.A_hat[0, 1], self.A_hat[1, 0]])`` gives me the same result as the original iterative method, please see my experiment here:

https://gist.github.com/shlff/02f5aac35f8793cdf08935e13174b131

Can you clarify a bit more?

oyamad:
@shlff Sorry you are right! `A_hat` is column-stochastic (not row-stochastic as in `P` in Lecture 25).

So the formula "inflow = outflow" here is: `A_hat[0, 1] * x[1] = A_hat[1, 0] * x[0]`.

oyamad:
2473f00dde9b62a8f017b34bd54de7a27a7b936b and d7d8f3af8b0e199d7d50f5634cacfd5fdd5cf86f should be reverted, sorry...

mmcky:
thanks @oyamad and @shlff

mmcky:
@jstac here is a preview

https://658133641d85b86841439172--nostalgic-wright-5fa355.netlify.app/lake_model.html

jstac:
Thanks @oyamad @mmcky @shlff .

- "steady state of the rate" -> "steady state rate"

- steady state rate needs to be defined

- "(which we discuss in section [Dynamics of an Individual Worker](link))" should be "using [a technique](same link) we previously introduced for computing stationary distributions of Markov chains"

@mmcky I'll step out of this now so please merge once these changes are made and you are happy (or request someone else to review).

mmcky:
> steady state rate needs to be defined

@shlff can I get your help in addressing this comment?

mmcky:
Latest preview is built: https://65822781c6711a007b89db11--nostalgic-wright-5fa355.netlify.app/lake_model.htm

shlff:
> > steady state rate needs to be defined
> 
> @shlff can I get your help in addressing this comment?
> 
> thanks for your review @jstac

Thanks @mmcky . I found the steady state rate has already been defined in the texts below the code:

https://github.com/QuantEcon/lecture-python.myst/blob/e585437300532cb7df82ac2dbb70744aba3ed213/lectures/lake_model.md?plain=1#L374-L380

I suggest we either leave it as it is or shift it up to the end of 

https://github.com/QuantEcon/lecture-python.myst/blob/e585437300532cb7df82ac2dbb70744aba3ed213/lectures/lake_model.md?plain=1#L198-L208

lectures/lake_model.md
```diff
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-1. provide a method to compute the steady state of the rate
+2. provide a method to compute the steady state rate using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
```
mmcky:
```suggestion
2. provide a method to compute the steady state vector of employment and unemployment rates $\bar x$ using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
```

mmcky: @shlff I have made an adjustment to include the notation $\bar x$ and be more specific with regards to vector of employment and unemployment rates. Do you agree?

shlff: Yep @mmcky , I think it is a good change. 

I just suggest shifting $\bar x$ to the end of ``... compute the steady state vector`` to avoid misunderstanding.

lectures/lake_model.md
```diff
@@ -218,7 +218,7 @@ This class will
 1. store the primitives $\alpha, \lambda, b, d$
 1. compute and store the implied objects $g, A, \hat A$
 1. provide methods to simulate dynamics of the stocks and rates
-1. provide a method to compute the steady state of the rate
+2. provide a method to compute the steady state vector of employment and unemployment rates $\bar x$ using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
```
mmcky:
```suggestion
2. provide a method to compute the steady state vector $\bar x$ of employment and unemployment rates using {ref}`a technique <dynamics_workers>` we previously introduced for computing stationary distributions of Markov chains
```

mmcky:
thanks for your comments @shlff and @jstac. I will merge this now. 

Thanks @mnshkw for your PR -- greatly appreciated.

PR #395, opened 2024-05-08
MAINT: update @jit(nopython=True) to @jit with numba>=0.59

mmcky:
This PR updates `njit` to use `jit` now that `numba>=0.59` and `nopython=True` is now the default behaviour of the `jit`

This resolves https://github.com/QuantEcon/meta/issues/112 for this lecture series


mmcky:
- [x] checked support files in `_static/lecture_specific`

mmcky:
There are a few speed regressions when applying these changes

Screenshot 2024-05-09 at 11 45 20 AM

@kp992 would you have any time to review this and see what might be causing the slowdown in execution speed?

kp992:
Checking this from numba repo:
https://github.com/numba/numba/blob/e15cd6870042125ff025e459f5f7558449b22cf8/numba/core/decorators.py#L49 -- Shows that nopython is set by default to True.

And `njit` is just `jit(nopython=True)`: https://github.com/numba/numba/blob/e15cd6870042125ff025e459f5f7558449b22cf8/numba/core/decorators.py#L250.

kp992:
Hmm, I guess there can 2 reasons behind this:
1. We are using GPU which calls cuda.jit internally in numba and the docs are just talking about CPU options.
2. Maybe the cache isn't updated and so the execution time has increased?

mmcky:
@kp992 I don't fully understand your comments. 

> We are using GPU which calls cuda.jit internally in numba and the docs are just talking about CPU options.

This PR is strictly update the Numba jit so I don't understand why GPU is involved?

> Maybe the cache isn't updated and so the execution time has increased?

For any lecture that has had code change then the cache should be invalidated and the lecture is run fresh. 

Perhaps we can link up and discuss this?

kp992:
Thanks @mmcky, I can try looking into it in more details and discuss on this.

kp992:
@mmcky Can we re-run and see if the time results still differ? Maybe sometimes we might have hit some outlier?

mmcky:
@kp992 there are still large differences:

Screenshot 2024-06-26 at 2 18 01 PM

kp992:
I will try to debug what's creating these differences.

mmcky:
- [ ] recheck results with latest anaconda environments to see if issue persists.

mmcky:
@kp992 and @HumphreyYang it looks like many of the large differences have resolved in the latest run and versions of `numba`.

kp992:
Thanks @kp992. This looks great and clean!

PR #399, opened 2024-08-04
ENH: Tom's edits for July

mmcky:
The current build cache has expired so waiting on #400


mmcky:
- [ ] re-run once the new cache is built on the `main` branch.

PR #402, opened 2024-08-07
ENH: Tom august updates to cass_coopmans_I

mmcky:


mmcky:
@shizejin it looks like there is an execution issue on this branch (not producing the plot)

```
T = paths[0].size - 1
```

is paths defined?

shizejin:
Thanks @mmcky. I tried to replicate the execution error on my local machine, but it builds well. Could you see more details about the execution issue from the server to better understand what's going wrong?

The `paths` should be have been created by the line `paths = shooting(pp, 0.2, 0.3, T=10)` in the cell above, and Tom and I didn't touch anything about the `shooting` function in this PR. Maybe there is something wrong when calling the `shooting` function, but I am sure what's causing it.

mmcky:
it appears there was a strange interaction with the `build cache` so a full build succeeds. Merging this now.

PR #410, opened 2024-08-17
ENH: updates from Tom and Humphrey

mmcky:


mmcky:
thanks @thomassargent30 and @HumphreyYang

PR #413, opened 2024-09-30
[odu] FIX: reformat wide equation

mmcky:
This PR reformats a wide equation in the ODU lecture


mmcky:
@jstac this PR fixes these long equations in the `odu` lecture. 

Screenshot 2024-10-02 at 12 44 50 pm

PR #418, opened 2024-11-04
Use float dtype for numpyro

### LINKED ISSUE #417: BUG: [bayes_nonconj] Execution error
@kp992 there is an execution error being reported in the `bayes_nonconj` lecture

```bash
TypeError[0m: Called multiply with a float0 array. float0s do not support any operations by design because they are not compatible with non-trivial vector spaces. No implicit dtype conversion is done. You can use np.zeros_like(arr, dtype=np.float) to cast a float0 array to a regular zeros array. 
If you didn't expect to get a float0 you might have accidentally taken a gradient with respect to an integer argument.
```

Do you have time to look into this and open a PR?

[bayes_nonconj.err.log](https://github.com/user-attachments/files/17613963/bayes_nonconj.err.log)

@kp992 also is there any way to make these log files more useful re: syntax?

kp992:
Use float `dtype` for numpyro backend in `bayes_nonconj`.

Fixes #417

### DIFF
--- a/lectures/bayes_nonconj.md
+++ b/lectures/bayes_nonconj.md
@@ -1,10 +1,10 @@
 ---
 jupytext:
   text_representation:
-    extension: .myst
+    extension: .md
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.13.8
+    jupytext_version: 1.16.4
 kernelspec:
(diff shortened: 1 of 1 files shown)


mmcky:
thanks so much @kp992

PR #427, opened 2024-12-04
MAINT: update publish to main

mmcky:


mmcky:
the link checker issue is fixed in #426

PR #439, opened 2025-01-11
FIX: syntax for strings python>=3.12

mmcky:
This PR fixes a syntax warning across the lecture series

@HumphreyYang if you have time would you mind review this PR to make sure an `r` shouldn't be an `rf` for example. It would be nice to have a double check of it.

### DIFF
--- a/lectures/ak2.md
+++ b/lectures/ak2.md
@@ -760,8 +760,8 @@ fig, axs = plt.subplots(3, 3, figsize=(14, 10))
 # quantities
 for i, name in enumerate(['K', 'Y', 'Cy', 'Co']):
     ax = axs[i//3, i%3]
-    ax.plot(range(T+1), quant_seq1[:T+1, i], label=name+', 1/3')
-    ax.plot(range(T+1), quant_seq2[:T+1, i], label=name+', 0.2')
+    ax.plot(range(T+1), quant_seq1[:T+1, i], label=f'{name}, 1/3')
+    ax.plot(range(T+1), quant_seq2[:T+1, i], label=f'{name}, 0.2')
     ax.hlines(init_ss[i], 0, T+1, color='r', linestyle='--')
     ax.legend()
     ax.set_xlabel('t')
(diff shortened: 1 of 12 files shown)


HumphreyYang:
Many thanks @mmcky. I will scan through the lectures.

HumphreyYang:
Hi Matt, 

I just pushed some other `r` cases I found in the lectures.

I think for this warning we need to look for any string that has a `\` that is not escaped : (

lectures/ak2.md
```diff
@@ -1179,26 +1179,26 @@ fig, axs = plt.subplots(3, 3, figsize=(14, 10))
 # quantities
 for i, name in enumerate(['K', 'Y', 'Cy', 'Co']):
     ax = axs[i//3, i%3]
-    ax.plot(range(T+1), quant_seq3[:T+1, i], label=name+', $\delta$s=0')
-    ax.plot(range(T+1), quant_seq4[:T+1, i], label=name+', $\delta$s=0.005')
+    ax.plot(range(T+1), quant_seq3[:T+1, i], label=rf'{name}, $\delta$s=0')
```
mmcky, line 1182: thanks @HumphreyYang good catch.

mmcky:
thanks @HumphreyYang I hadn't searched for the embedded case. Nice catch.

jstac:
Thanks @mmcky and @HumphreyYang for the fixes.

Once you get this working please roll out to the other lecture series.

mmcky:
@HumphreyYang once a `r` string is setup we need to remove any `\\` so there are quite a few cases to think about :-).

mmcky:
@HumphreyYang on my local I am getting Syntax Warnings for docstrings. Is this happening to you?

For example from `lake_model` this line

```
Finds the steady state of the system :math:`x_{t+1} = \hat A x_{t}`
```

in a docstring is causing issues with `\hat`.

HumphreyYang:
Thanks @mmcky,

I did not get this error message. This is interesting, perhaps we need to escape things in the docstring?

mmcky:
> Thanks @mmcky,
> 
> I did not get this error message. This is interesting, perhaps we need to escape things in the docstring?

thanks for letting me know @HumphreyYang. It's an odd error. It shouldn't be causing issues as docstrings should be able to contain latex markup. I'll have to do some further digging to see if I can replicate this in another environment.

Issue #452, opened 2025-05-08
[GPU] disable install of GPU related software + add standard GPU admonition


mmcky:
Update lectures installs based our approach to notifying instead of installing

- [x] we should still disable jax, torch, pryo-ppl and numpyro installs as these are hardware dependent and need to be maintained carefully. Add the lecture notice on GPU instead.
- [x] add in the standard GPU admonition for users to install the software needed. 
- [ ] update QuantEcon.manual

PR #453, opened 2025-05-08
MAINT: remove jax, pyro, torch, and gpu related software installs + GPU admonition

mmcky:
This PR addresses #452

~~This PR currently breaks `collab` testing. While google `collab` has `jax` it doesn't have a few other packages related to GPU such as `pyro` which is causing 1-2 lectures to fail. The installs will be listed at the top of the lecture, however a user will need to run that cell to install those packages for it to work.~~ 🤔 

I have switched these back on for local execution (this should fix `collab` while not upgrading the software for the other actions). I have documented this complexity and issue in #454 so doesn't need to be fixed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
         shell: bash -l {0}
         run: |
           pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu128
+          pip install pyro-ppl
           pip install --upgrade "jax[cuda12-local]"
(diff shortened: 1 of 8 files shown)


mmcky:
@jstac this PR is adding a general `GPU` admonition to lectures that make use of GPU processing, in similar style to JAX. 

https://6822f27dd8e5c4e161464ee6--nostalgic-wright-5fa355.netlify.app/ar1_bayes.html

I have also shuffled a few installs around so that GPU related installs are separate to general packages.

PR #465, opened 2025-06-17
[opt_transport] Update Description on outputs of LP solvers

### LINKED ISSUE #461: [opt_transport] Update the descriptions on outputs of LP solvers
The following is the output from ChatGPT (with a few edits by myself). (I gave these three issues and asked to generate an Issue draft.)

===

The file `lectures/opt_transport.md` contains several outdated or incorrect statements about the LP solvers:

1. **Obsolete SciPy solver description**  
   The lecture still refers to the legacy (revised) simplex implementation in `scipy.optimize.linprog`. Since SciPy 1.6.0 (February 2021), the default backend is HiGHS. All references to the “old” solver, its warnings, and performance characteristics should be updated to reflect HiGHS.

2. **Incorrect algorithm comparison**  
   > “Since the two LP solvers use the same simplex algorithm, we expect to get exactly the same solutions.”  

   This is incorrect:  
   - `quantecon.optimize.linprog_simplex` and SciPy’s `linprog` (HiGHS backend) use **different underlying algorithms**.  

3. **Redundant dual-problem solve**  
   `linprog_simplex` already computes and returns dual variables alongside the primal solution. The separate dual-LP call is unnecessary. Consider removing the extra dual solve and showing how to extract shadow prices directly from the `linprog_simplex` result.

===

HumphreyYang:
This PR fixes #461.

### DIFF
--- a/lectures/opt_transport.md
+++ b/lectures/opt_transport.md
@@ -4,7 +4,7 @@ jupytext:
     extension: .md
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.14.4
+    jupytext_version: 1.17.1
 kernelspec:
   display_name: Python 3 (ipykernel)
   language: python
@@ -330,12 +330,12 @@ Let's write Python code that sets up the problem and solves it.
 m = 3
 n = 5
 
-p = np.array([50, 100, 150])
-q = np.array([25, 115, 60, 30, 70])
+p = np.array([50.0, 100.0, 150.0])
+q = np.array([25.0, 115.0, 60.0, 30.0, 70.0])
 
-C = np.array([[10, 15, 20, 20, 40],
-              [20, 40, 15, 30, 30],
-              [30, 35, 40, 55, 25]])
+C = np.array([[10.0, 15.0, 20.0, 20.0, 40.0],
+              [20.0, 40.0, 15.0, 30.0, 30.0],
+              [30.0, 35.0, 40.0, 55.0, 25.0]])
 
 # Vectorize matrix C
 C_vec = C.reshape((m*n, 1), order='F')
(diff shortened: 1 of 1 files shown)


HumphreyYang:
Hi @mmcky,

I think this is ready for your review. 

I tried to enforce both functions to use simplex method but the `sympy` gives deprecation warning.

Please let me know if there is anything I should update!

mmcky:
thanks @HumphreyYang -- I'll review this soon.

lectures/opt_transport.md
```diff
@@ -502,7 +508,13 @@ minimization, we need to put a negative sign before vector `c`.
 res_qe = linprog_simplex(-c, A_eq=A_eq, b_eq=b_eq)
 ```
 
-Since the two LP solvers use the same simplex algorithm, we expect to get exactly the same solutions
+While the two LP solvers use different algorithms (HiGHS vs. simplex), both should find optimal solutions.
+
+The solutions differs since there are multiple optimal solutions, but the objective values are the same
+
+```{code-cell} ipython3
+np.allclose(- res_qe.fun, res.fun)
```
mmcky: @HumphreyYang I am not familiar with this syntax `-` is that the negative of `res_qe.fun`

Just not sure why there is a space.

mmcky (commented):
thanks @HumphreyYang this looks great. Just had one question.

lectures/opt_transport.md
HumphreyYang: Yes it is the negative of the objective of the quantecon result.

I will remove the space!

HumphreyYang:
Many thanks @mmcky, I just removed the space!

mmcky:
Thanks for your work on this @HumphreyYang -- its great.

Issue #463, opened 2025-06-16
`np.sum(a * b)` versus `a @ b`


oyamad:
Inner dots are computed as `np.sum(a * b)` in many places. Consider replacing it with `a @ b`. (See also the 3rd item in https://github.com/QuantEcon/lecture-python.myst/issues/169#issue-946954733.)

Below is the output of ChatGPT:

===
- **Under the hood**: `a @ b` (or `np.dot(a, b)`) delegates to optimized BLAS libraries that implement fused multiply-add (FMA) operations, eliminating the temporary array and reducing rounding errors compared to separate multiply and sum steps. (https://stackoverflow.com/questions/18092984/numpy-difference-between-dota-b-and-ab-sum)  
- **Performance overhead**: `np.sum(a * b)` allocates an intermediate array and then sums it, incurring extra memory traffic and Python-level looping; by contrast, BLAS kernels stream data efficiently in optimized C code. (https://stackoverflow.com/questions/75556221/why-is-np-dot-so-much-faster-than-np-sum)  
- **Numerical accuracy**: BLAS libraries use hardware FMA instructions to fuse multiply and add into a single rounding operation, improving precision over two separate operations. (https://stackoverflow.com/questions/45200278/numpy-fusing-multiply-and-add-to-avoid-wasting-memory)  
- **Official recommendation**: NumPy’s documentation advises using `np.dot(a, b)` or the `@` operator (via `np.matmul`, per PEP 465) for dot products and matrix multiplication. (https://numpy.org/doc/stable/reference/generated/numpy.dot.html, https://numpy.org/doc/stable/reference/generated/numpy.matmul.html, https://peps.python.org/pep-0465/)  

===

jstac:
Thanks @oyamad , this is very helpful -- and interesting.

@mmcky -- would you mind to supervise fixes?  Perhaps @oyamad 's students or one of our RAs can take this job?

mmcky:
These are the following instances using a regex pattern

- [x] lectures/aiyagari.md:    return np.sum(asset_probs * am.a_vals) https://github.com/QuantEcon/lecture-python.myst/pull/467
- [ ] lectures/career.md:        self.F_mean = np.sum(self.θ * self.F_probs)
- [ ] lectures/career.md:        self.G_mean = np.sum(self.ϵ * self.G_probs)
- [ ] lectures/lake_model.md:                    β * γ * np.sum(np.maximum(U, V) * p_vec)
[...]

PR #467, opened 2025-06-19
[aiyagari] MAINT: transfer np.sum(a * b) to a @ b

mmcky:
This PR fixes part of #463 

- [x] apply fix and compare to original code
- [x] remove original code and comparison code

### DIFF
--- a/lectures/aiyagari.md
+++ b/lectures/aiyagari.md
@@ -368,86 +368,6 @@ The following code draws aggregate supply and demand curves.
 The intersection gives equilibrium interest rates and capital.
 
 ```{code-cell} ipython3
-:tags: [hide-input]
-
-A = 1.0
-N = 1.0
-α = 0.33
-β = 0.96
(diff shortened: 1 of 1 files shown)


mmcky:
@HumphreyYang this is an easy one, but would you mind reading through https://github.com/QuantEcon/lecture-python.myst/issues/463#issuecomment-2986383213 and letting me know if you approve of this change to:

- [x] verify change is correct
- [x] check the assert between `original code` and `new code`

HumphreyYang (approved):
Hi @mmcky, this looks great to me.

I think we can go ahead with a global switch, as I don't expect this change to cause much trouble.

After that, we can review all the updated lectures to ensure they’re consistent with the previous results!

mmcky:
thanks @HumphreyYang

mmcky:
thanks @HumphreyYang -- we are getting some help from RAs in Japan as well.

Issue #457, opened 2025-05-20
[cass_koopmans_1] Turnpike property


oyamad:
Shouldn't the "[Turnpike Property](https://python.quantecon.org/cass_koopmans_1.html#a-turnpike-property)" refer also to the independence of the initial condition? I.e., if T is large enough, then independent of the initial K_0, the optimal path of K_t stays close to the steady state value.

The code of `plot_saving_rate` should be extended to accept a list of `k0`s, to compare the trajectories from different `k0`s.

mmcky:
@oyamad I know very little about this lecture but I thought I would see what `Claude Sonnet 4` model would say as a fix to this issue by reading the Cass Koopmans 1 lecture. 

@HumphreyYang would you have any time to take a look and let me know how useful this feedback is. 

Here is what it said.

---

# Fix for Cass-Koopmans Lecture - Turnpike Property Section

## Issue Summary
The GitHub issue points out that the Turnpike Property section should demonstrate that the property holds **regardless of the initial capital condition K₀**. Currently, the lecture only shows convergence from one starting point, but the turnpike property is stronger - it should hold for multiple different initial conditions.

## Proposed Changes

### 1. Enhanced `plot_saving_rate` function

Replace the existing `plot_saving_rate` function with this enhanced version that accepts multiple initial capital values:

```python
def plot_saving_rate(pp, c0, k0_arr, T_arr, k_ter=0, k_ss=None, s_ss=None):
    """
    Enhanced saving rate plotting function that accepts either:
    - A single k0 value (backward compatibility)
    - A list of k0 values to demonstrate turnpike property
    """
    # Handle backward compatibility - if k0_arr is a scalar, convert to list
    if np.isscalar(k0_arr):
        k0_arr = [k0_arr]
        single_k0 = True
    else:
        single_k0 = False
    
    if single_k0:
        # Original behavior for single k0
        fig, axs = plt.subplots(2, 2, figsize=(12, 9))
        c_paths, k_paths = plot_paths(pp, c0, k0_arr[0], T_arr, k_ter=k_ter, k_ss=k_ss, axs=axs.flatten())
        for i, T in enumerate(T_arr):
            s_path = saving_rate(pp, c_paths[i], k_paths[i])
            axs[1, 1].plot(s_path)
        axs[1, 1].set(xlabel='t', ylabel='$s_t$', title='Saving rate')
        if s_ss is not None:
[...]

PR #469, opened 2025-06-19
[cass_koopmans_1] Claude review of issue #457

mmcky:
This PR implements the suggestions by `Claude` LLM for #457 

**Notes:** This has been really interesting. The suggestions by `Claude` don't take into account the flow of the program -- in Jupyter notebooks context. For example, the suggestions to update the section, come before the function `plot_saving_rate` is defined causing an execution error. It also makes some assumptions that `s_ss` is a defined variable by passing it in -- but it is never used in the lecture. 

Also its solution to plotting is quite verbose, when it basically replicates the same diagram as found in the section before it.

### DIFF
--- a/lectures/cass_koopmans_1.md
+++ b/lectures/cass_koopmans_1.md
@@ -782,7 +782,24 @@ the optimal capital stock stays close to
 its steady state value most of the time.
 
 ```{code-cell} ipython3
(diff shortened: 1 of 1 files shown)


mmcky:
closing as no longer needed.

PR #471, opened 2025-06-23
[mccall_model] Replace `np.sum(a * b)` with `a @ b`

UdohNakamura:
Replace `np.sum(a * b)` with `a @ b`
#463

### DIFF
--- a/lectures/mccall_model.md
+++ b/lectures/mccall_model.md
@@ -404,7 +404,7 @@ class McCallModel:
         # Evaluate value for each state-action pair
         # Consider action = accept or reject the current offer
         accept = w[i] / (1 - β)
-        reject = c + β * np.sum(v * q)
+        reject = c + β * (v @ q)
 
         return np.array([accept, reject])
 ```
@@ -488,7 +488,7 @@ def compute_reservation_wage(mcm,
 
     # == Now compute the reservation wage == #
 
-    return (1 - β) * (c + β * np.sum(v * q))
+    return (1 - β) * (c + β * (v @ q))
 ```
 
 The next line computes the reservation wage at  default parameters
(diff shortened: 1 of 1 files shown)


mmcky:
thank you @UdohNakamura for this FIX - I have enabled the tests to run.

lectures/mccall_model.md
```diff
@@ -404,7 +404,7 @@ class McCallModel:
         # Evaluate value for each state-action pair
         # Consider action = accept or reject the current offer
         accept = w[i] / (1 - β)
-        reject = c + β * np.sum(v * q)
+        reject = c + β * (v @ q)
```
HumphreyYang, line 407: Many thanks @UdohNakamura and @mmcky,

When building on my side I got
```
<string>:3: NumbaPerformanceWarning: '@' is faster on contiguous arrays, called on (Array(float64, 1, 'C', False, aligned=True), Array(float64, 1, 'A', False, aligned=True))
```

Upon checking, I found we annotate the type 

```
    ('w', float64[:]),   # array of wage values, w[i] = wage at state i
    ('q', float64[:])    # array of probabilities
```

The following type will help (see more [here](https://stackoverflow.com/questions/77870893/numbaperformancewarning-is-faster-on-contiguous-arrays-warning-on-jitcl?utm_source=chatgpt.com):

```
    ('w', float64[::1]),   # array of wage values, w[i] = wage at state i
    ('q', float64[::1])    # array of probabilities
```

Please let me know what you think.

oyamad: @HumphreyYang Thanks, I didn't know that. Maybe write `float64[::1]` in place of `float64[:]` and explain that it is needed to have `@` work inside a `jitclass`?

mmcky: thanks @HumphreyYang and @oyamad 

@HumphreyYang I will merge these great PRs and then you can do a pass to fix this.

mmcky (approved):

mmcky:
thanks so much of this PR @UdohNakamura

PR #472, opened 2025-06-25
Replace np.sum(a * b) with a @ b in lecture-python.myst.mccall_q.md

suda-yuga:
Replace np.sum(a * b) with a @ b


mmcky (approved):

mmcky:
thanks so much for this PR @suda-yuga

PR #474, opened 2025-06-25
[multivariate_normal] Replace np.sum(a * b) with a @ b

suda-yuga:
Replace np.sum(a * b) with a @ b


mmcky:
thanks for the PR @suda-yuga

PR #475, opened 2025-06-25
[opt_transport] Replace np.sum(a * b) with np.vdot(a, b)

suda-yuga:
Replace np.sum(a * b) with np.vdot(a, b)

### DIFF
--- a/lectures/opt_transport.md
+++ b/lectures/opt_transport.md
@@ -685,6 +685,8 @@ total_cost = np.vdot(X, C)
 total_cost
 ```
 
+Here we use [np.vdot](https://numpy.org/doc/stable/reference/generated/numpy.vdot.html) for the trace inner product of X and C
+
 ### A Larger Application
 
 Now let's try using the same package on a slightly larger application.
--- a/lectures/opt_transport.md
+++ b/lectures/opt_transport.md
@@ -681,7 +681,7 @@ X
 Sure enough, we have the same solution and the same cost
 
 ```{code-cell} ipython3
-total_cost = np.sum(X * C)
+total_cost = np.vdot(X, C)
 total_cost
 ```
 


HumphreyYang:
Hi @suda-yuga,

Thanks so much for your PR! This is a nice application of `vdot`. I am  curious would `np.sum(a*b)` be more transparent than `np.vdot(a, b)` in this case as we are summing over a 2-D matrix?

suda-yuga:
> Hi @suda-yuga,
> 
> Thanks so much for your PR! This is a nice application of `vdot`. I am curious would `np.sum(a*b)` be more transparent than `np.vdot(a, b)` in this case as we are summing over a 2-D matrix?

Thank you for your comment.

You're absolutely right — in this case, np.sum(a * b) is arguably more transparent and easier to read when working with 2-D matrices.

The motivation behind this change was based on prior discussions (https://github.com/QuantEcon/lecture-python.myst/issues/463) suggesting that a @ b or np.vdot(a, b) can be more efficient than np.sum(a * b) due to avoiding intermediate arrays and leveraging BLAS-level optimizations. I applied the same reasoning to the Frobenius inner product of matrices here.

That said, I agree with you — clarity matters, and np.sum(a * b) may better communicate the intent in this specific context. I'm happy to revert or revise the change depending on what's preferred for readability.

HumphreyYang:
Thanks @suda-yuga! I'm happy either way. I just feel that it's less obvious to readers what we're trying to compute here, unlike the more straightforward cases where we have two 1-D arrays. For the sake of presentation, I'm inclined to close this PR—though I really like what you've done!

Hi @mmcky, I will leave it to you to make the call!

oyamad:
Maybe write something like "Here we use `np.vdot` (with a link to https://numpy.org/doc/stable/reference/generated/numpy.vdot.html) (instead of `np.dot` or `@`), where `X` and `C` are 2-D arrays"?

oyamad:
Or "Here we use `np.vdot` for the trace inner product of `X` and `C`".
(The notation "tr (C' X)" is already introduced (with no explanation).)

mmcky:
Thanks @oyamad for the suggestion. 

I think being explicit around the use of `vdot` is a good idea, specifying what it achieves. I have added your second suggestion with a link to the `vdot` documentation.

mmcky:
thanks @HumphreyYang and @oyamad for your valuable feedback and review. 

Thanks @suda-yuga for the PR.

PR #482, opened 2025-07-08
[kalman_2] Remove redundant axhline plotting

suda-yuga:
Remove redundant ax.axhline call outside the simulation loop.

- The true effort level (u_0) lines are already plotted inside the simulate_workers function.
- The external ax.axhline call used u_0, which was undefined outside the loop.
- This cleanup avoids redundant code and potential NameError.


mmcky:
thanks for this PR @suda-yuga

PR #486, opened 2025-07-14
[likelihood_ratio_process] Update the lecture with new simulations

HumphreyYang:
This PR updates the lecture `likelihood_ratio_process` with new examples and simulations.

It also includes minor edits to `prob_matrix`.


mmcky:
thanks @HumphreyYang

PR #501, opened 2025-07-29
job search update

jstac:
Changing job search with separation to use JAX

@longye-tian @HumphreyYang 

Could one of you please review this?  The last figure looks a bit different.


longye-tian:
Hi John, @jstac ,

I just compare the numba version and the jax version.

I find that Jax version is correct, and numba version has a small problem.

In the original numba code, we didn't put everything in functions, so we have in the first loop over unemployment compensation c from 2 to 12.

So the model in the numba version when looping over discount factor and separation rate, we have c=12 instead of the default value of 6. 

Similar happens when we loop discount factor openly without the function, but since the last loop is 0.98 over 0.99 the difference is not too big.

**So the conclusion is that the Jax version is correct and better.**

Best,
Longye

lectures/mccall_model_with_separation.md
```diff
@@ -519,57 +503,54 @@ alpha_vals = np.linspace(0.05, 0.5, grid_size) # separation rate
...
-w_bar_vals = np.empty_like(c_vals)
+```{code-cell} ipython
 
-fig, ax = plt.subplots()
+def compute_res_wage_given_c(c):
+    model = Model(c=c)
+    w_bar = compute_reservation_wage(model)
+    return w_bar
 
-for i, c in enumerate(c_vals):
```
longye-tian, line 529: we didn't put this part in the function.

So later one when we loop over discount factor, we have c=12 which is the value from the last loop

lectures/mccall_model_with_separation.md
```diff
@@ -519,57 +503,54 @@ alpha_vals = np.linspace(0.05, 0.5, grid_size) # separation rate
...
 
-```{code-cell} python3
-fig, ax = plt.subplots()
+```{code-cell} ipython
+def compute_res_wage_given_beta(β):
+    model = Model(β=β)
+    w_bar = compute_reservation_wage(model)
+    return w_bar
 
-for i, β in enumerate(beta_vals):
```
longye-tian, line 547: Similar in here

lectures/mccall_model_with_separation.md
```diff
@@ -519,57 +503,54 @@ alpha_vals = np.linspace(0.05, 0.5, grid_size) # separation rate
...
-```{code-cell} python3
-fig, ax = plt.subplots()
+```{code-cell} ipython
+
+def compute_res_wage_given_alpha(α):
+    model = Model(α=α)
+    w_bar = compute_reservation_wage(model)
+    return w_bar
 
-for i, α in enumerate(alpha_vals):
```
longye-tian, line 564: similar in here

longye-tian (approved):
Jax version is correct

HumphreyYang:
Many thanks @jstac and @longye-tian,

I agree. One way to verify that this is the issue is to add `mcm = McCallModel()` before we run the experiment at each cell so it refresh the parameter that is carried over from the previous experiment.

I got the same solution as JAX once we do that:
https://gist.github.com/a8b53da3d0a4fbc2956bc29e1d000013.git

mmcky:
@longye-tian and @HumphreyYang outstanding work. 

- [ ] ~~open a companion PR to fix `numba` implementation.~~

HumphreyYang:
Thanks @mmcky,

Would you like me to open a PR that fix the numba code? Just curious if we are replacing the numba code with JAX code.

mmcky:
> Thanks @mmcky,
> 
> Would you like me to open a PR that fix the numba code? Just curious if we are replacing the numba code with JAX code.

Ah. Thank you @HumphreyYang. I am rushing around the repos at the moment. You're 100% correct. This is replacing the `numba` code (I thought from the discussion we had both versions). Nothing to do here :-)

mmcky:
Thanks @HumphreyYang and @longye-tian for your helpful reviews. 

@jstac just checking do you intend to make any other changes before merge?

jstac:
Many thanks @longye-tian for the careful review, it's much appreciated.  Thanks also @HumphreyYang for confirming and @mmcky for supervising.

As @HumphreyYang says, we only need the JAX version. Can someone please do one more overall sanity check and make sure everything makes sense.  Then please flag it for merging.  I don't need to see it again.

HumphreyYang:
Many thanks @jstac and @mmcky,

I will update the static figures in the lecture to the new version and change a few variable names to unicode. Other than those, it looks good to me!

PR #529, opened 2025-08-12
CI: add action to check code warnings

mmcky:
This PR enables a custom QuantEcon GitHub Action ([check-warnings](https://github.com/QuantEcon/meta/blob/main/.github/actions/check-warnings/README.md)) for automated checking of the following warnings contained in code output:

These include checking for [all `python` warnings](https://github.com/QuantEcon/meta/issues/201)

and the action is specifically designed to parse `Jupyter Book` html output.


mmcky:
This failed due to `FutureWarning` being in the text of `kesten_processes` so it is a false positive. 

Try to improve the checker to be contained within code output only. 

Perhaps the checker could work on a collection of notebooks (for more well defined checks of `code-output` in json rather than through `html`

mmcky:
💡 I wonder if instead of opening an Issue the report is added to the PR thread instead so it is fixed before the PR is merged. That might be a better workflow for these SyntaxWarnings. (although sometimes they are caused upstream and that might make merging a PR problematic).

mmcky:
@HumphreyYang @jstac I have made an automation to check for `SyntaxWarning`, `DeprecationWarning` and `FutureWarning` in all code outputs. 

We have a few options for how to use it. 

It is currently setup to `fail` a PR if they are found, but post to the PR feed (automatically) with information about where it is to fix like: https://github.com/QuantEcon/lecture-python.myst/pull/529#issuecomment-3181914691

The only issue with this behaviour is that if a package is updated (for example) causing a warning this workflow may pickup a failure in an unrelated lecture to the PR. 

We also have the option to `not` fail the workflow but report the information as an `Issue` (automatically) like this: https://github.com/QuantEcon/lecture-python.myst/issues/535

I think I am leaning towards the opening an `Issue` approach to prevent any holdups and confusion in the PR process.  However, the `pro` for failing the PR is that they never make it into the live site (at the `con` of fixing unrelated errors from time to time in the PR). 

Any thoughts?

jstac:
Very cool, thanks @mmcky ! 

I'm +1 for opening an issue -- but it might be good if one or both of us gets pinged, since these problems need to be fixed asap.  Is there an easy way to do that by tweaking the output (e.g.: add 'CC someone')

HumphreyYang:
Hi @mmcky,

This is super cool! +1 for issues and I can also be pinned as a repairer. I recall that we also have some runtime warning that is intentional in the advanced series. For example:

```
/home/runner/miniconda3/envs/quantecon/lib/python3.12/site-packages/quantecon/_compute_fp.py:152: RuntimeWarning: max_iter attained before convergence in compute_fixed_point
  warnings.warn(_non_convergence_msg, RuntimeWarning)
```
  
Should we migrate this to printing instead of `RuntimeWarning`?

mmcky:
@jstac @HumphreyYang made some changes for the best of both worlds. Now the `create-issue` mode will also post to the PR thread to notify it has picked up some warnings (along with a link to the issue that was opened) -- but the action won't fail the PR run. 

Doing last test run now, and then you shouldn't get pinged any more :-)

mmcky:
> I recall that we also have some runtime warning that is intentional in the advanced series. For example:
> 
> ```
> /home/runner/miniconda3/envs/quantecon/lib/python3.12/site-packages/quantecon/_compute_fp.py:152: RuntimeWarning: max_iter attained before convergence in compute_fixed_point
>   warnings.warn(_non_convergence_msg, RuntimeWarning)
> ```
> 
> Should we migrate this to printing instead of `RuntimeWarning`?

Thanks @HumphreyYang that is an interesting use case of RuntimeWarning. Currently we are only looking for `DeprecationWarning`, `SyntaxWarning`, and `FutureWarning` so this wouldn't be picked up at the moment, but that means all RuntimeWarnings will be ignored. 🤔 

The warnings we are looking for are configurable in the action https://github.com/QuantEcon/meta/pull/185#issue-3302813933

HumphreyYang:
> > I recall that we also have some runtime warning that is intentional in the advanced series. For example:
> > ```
> > /home/runner/miniconda3/envs/quantecon/lib/python3.12/site-packages/quantecon/_compute_fp.py:152: RuntimeWarning: max_iter attained before convergence in compute_fixed_point
> >   warnings.warn(_non_convergence_msg, RuntimeWarning)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should we migrate this to printing instead of `RuntimeWarning`?
> 
> Thanks @HumphreyYang that is an interesting use case of RuntimeWarning. Currently we are only looking for `DeprecationWarning`, `SyntaxWarning`, and `FutureWarning` so this wouldn't be picked up at the moment, but that means all RuntimeWarnings will be ignored. 🤔
> 
> The warnings we are looking for are configurable in the action [QuantEcon/meta#185 (comment)](https://github.com/QuantEcon/meta/pull/185#issue-3302813933)

Thanks @mmcky! just curious if we should avoid using warnings when we update lectures : )

mmcky:
@HumphreyYang this is working well. 

It has reported some actual issues in https://github.com/QuantEcon/lecture-python.myst/issues/548

mmcky:
@copilot is there a way to run the `check-warnings` action only once when the PR is opened for the first time. This would enable us to collect the data, but not open repeated issues due to updates to the PR via additional commits.

mmcky:
@copilot  is there a way to run the check-warnings action only once when the PR is opened for the first time. This would enable us to collect the data, but not open repeated issues due to updates to the PR via additional commits.

.github/workflows/ci.yml
```diff
@@ -72,6 +75,14 @@ jobs:
         shell: bash -l {0}
         run: |
           jb build lectures --path-output ./ -n -W --keep-going
+      - name: Check for Python warnings
+        uses: QuantEcon/meta/.github/actions/check-warnings@main
```
mmcky:
```suggestion
        uses: QuantEcon/actions-check-warnings@v1
```

.github/workflows/ci.yml
```diff
@@ -79,6 +82,14 @@ jobs:
         shell: bash -l {0}
         run: |
           jb build lectures --path-output ./ -n -W --keep-going
+      - name: Check for Python warnings
+        uses: QuantEcon/actions-check-warnings@v1
```
mmcky:
```suggestion
        uses: QuantEcon/action-check-warnings@v1
```

.github/workflows/ci.yml
```diff
@@ -79,6 +82,14 @@ jobs:
         shell: bash -l {0}
         run: |
           jb build lectures --path-output ./ -n -W --keep-going
+      - name: Check for Python warnings
+        uses: QuantEcon/action-check-warnings@v1
```
mmcky:
```suggestion
        uses: QuantEcon/action-check-warnings@v1.0.0
```

mmcky:
Closing as will continue development and build a test repository.

Issue #478, opened 2025-07-01
von Neumann lecture improvements


jstac:
- Only part of `class Neumann(object):` is showing --- show all
- the word `object` should be removed from the above
- use correct theorem and proof environments (ask @mmcky)

PR #533, opened 2025-08-12
[von_neumann_model] lecture improvements

xuanguang-li:
Key changes
- Set the max-height of cells to none at the beginning
- Removed `object` after `class Nuemann`
- Adopted `sphinx-proof` environments

Question
- As I used the `sphinx-proof`, assumptions numbered as "I, II"  have been changed to "1, 2". But in the definition of `class Neumann`, some variables are named as `AI` and `AII`. Should I change the names of these variables to `A1` and `A2`? @mmcky

For example,
```
def __init__(self, A, B):

        self.A, self.B = list(map(self.convert, (A, B)))
        self.m, self.n = self.A.shape

        # Check if (A, B) satisfy the basic assumptions
        assert self.A.shape == self.B.shape, 'The input and output matrices \
              must have the same dimensions!'
        assert (self.A >= 0).all() and (self.B >= 0).all(), 'The input and \
              output matrices must have only non-negative entries!'

        # (1) Check whether Assumption 1 is satisfied:
        if (np.sum(B, 0) <= 0).any():
            self.AI = False
        else:
            self.AI = True

        # (2) Check whether Assumption 2 is satisfied:
        if (np.sum(A, 1) <= 0).any():
            self.AII = False
        else:
            self.AII = True
```

linked to #478

### DIFF
--- a/lectures/von_neumann_model.md
+++ b/lectures/von_neumann_model.md
@@ -61,7 +61,6 @@ np.set_printoptions(precision=2)
 The code below provides the `Neumann` class
 
 ```{code-cell} ipython3
-:tags: [collapse-20]
 
 class Neumann:
 
--- a/lectures/von_neumann_model.md
+++ b/lectures/von_neumann_model.md
@@ -19,14 +19,6 @@ kernelspec:
         </a>
 </div>
 
-<style>
-/* Only expand code cells on this page */
-div.cell div.highlight,
-div.cell div.output_scroll {
-    max-height: none !important;
-    overflow-y: visible !important;
-}
-</style>
 ```
 
 ```{index} single: python
@@ -93,9 +85,9 @@ class Neumann:
         Ap - the cost of each activity
 
     Both A and B have non-negative entries. Moreover, we assume that
-    (1) Assumption 1 (every good which is consumed is also produced):
+    (1) Assumption I (every good which is consumed is also produced):
         for all j, b_{.,j} > 0, i.e. at least one entry is strictly positive
-    (2) Assumption 2 (no free lunch):
+    (2) Assumption II (no free lunch):
         for all i, a_{i,.} > 0, i.e. at least one entry is strictly positive
 
     Parameters
@@ -139,9 +131,9 @@ class Neumann:
           - number of goods          : {n}
           - number of activities     : {m}
 
-        Assumptions:// TODO:change {AI} variable into A1?
-          - A1:  every column of B has a positive entry    : {AI} 
-          - A2: every row of A has a positive entry       : {AII}
+        Assumptions://
+          - AI:  every column of B has a positive entry    : {AI} 
+          - AII: every row of A has a positive entry       : {AII}
 
         """
         # Irreducible                                       : {irr}
@@ -382,7 +374,7 @@ an economy.
 
 Two key assumptions restrict economy $(A,B)$:
 
-````{prf:assumption} every good that is consumed is also produced)
+````{prf:assumption} every good that is consumed is also produced
 :label: assumption1
 
 ```{math}
--- a/lectures/von_neumann_model.md
+++ b/lectures/von_neumann_model.md
@@ -109,13 +109,13 @@ class Neumann:
         assert (self.A >= 0).all() and (self.B >= 0).all(), 'The input and \
               output matrices must have only non-negative entries!'
 
-        # (1) Check whether Assumption 1 is satisfied:
+        # (1) Check whether Assumption I is satisfied:
         if (np.sum(B, 0) <= 0).any():
             self.AI = False
         else:
             self.AI = True
 
-        # (2) Check whether Assumption 2 is satisfied:
+        # (2) Check whether Assumption II is satisfied:
         if (np.sum(A, 1) <= 0).any():
             self.AII = False
         else:
@@ -328,13 +328,13 @@ We use the following notation.
 $\mathbf{0}$ denotes
 a vector of zeros.
 
-We call an $n$-vector  positive and write
+We call an $n$-vector **positive** and write
 $x\gg \mathbf{0}$ if $x_i>0$ for all $i=1,2,\dots,n$.
 
-We call a vector  non-negative and write $x\geq \mathbf{0}$ if $x_i\geq 0$ for
+We call a vector **non-negative** and write $x\geq \mathbf{0}$ if $x_i\geq 0$ for
 all $i=1,2,\dots,n$.
 
-We call a vector  semi-positive and written $x > \mathbf{0}$ if
+We call a vector **semi-positive** and written $x > \mathbf{0}$ if
 $x\geq \mathbf{0}$ and $x\neq \mathbf{0}$.
 
 For two conformable vectors $x$ and $y$, $x\gg y$,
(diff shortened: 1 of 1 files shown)


mmcky:
Thanks for the PR @xuanguang-li. I will take a look tomorrow.

mmcky:
@jstac -- @xuanguang-li has asked a really nice style clarification question.

In the case where there are two or more conditions are states, should a python variable name always use numbers such as `A1`, `A2`? 

What makes me pause is that in some documents `Assumption` admonitions can be roman numerals in style and in fact the docstring associated with this code uses roman numerals so it may be "closer to the description" keeping the roman numerals in the code. 

```
    Both A and B have non-negative entries. Moreover, we assume that
    (1) Assumption I (every good which is consumed is also produced):
        for all j, b_{.,j} > 0, i.e. at least one entry is strictly positive
    (2) Assumption II (no free lunch):
        for all i, a_{i,.} > 0, i.e. at least one entry is strictly positive
```

jstac:
> so it may be "closer to the description" keeping the roman numerals in the code.

That sounds like good reasoning to me @mmcky .

xuanguang-li:
Hi John and @mmcky,
 
Thanks for your feedback. 

I understand your point that I should keep the variable names consistent with the docstrings. Should I still use `sphinx-proof` directives to write assumptions?
```
```{prf:assumption} every good that is consumed is also produced
:label: assumption1

$$
  b_{.,j} > \mathbf{0}\hspace{5mm}\forall j=1,2,\dots,n
$$
```
If I use `sphinx-proof` directives, `Assumption` admonitions will be numbered as `Assumption 1` rather than in Roman numerals,  as shown in the attached pictures. The reference of this assumption using `{prf: ref}` will also display as `Assumption 1`.

image
image

mmcky:
thanks @xuanguang-li let me think this through, I take your point that we will have mixed systems in the lecture 👍 

One option would be to enable roman numeral styles for `sphinx-proof`.

xuanguang-li:
Thanks for your explanation. I kept the Roman numerals in the code while using the default `sphinx-proof` setting, as enabling Roman numerals requires editing `conf.py`.

Regarding the suggestion, "Only part of class Neumann(object): is showing --- show all", I tried adding a `<style>` tag in this file and creating a `custom.css` file to remove the cell height restriction, but both attempts failed. It seems I need to edit the `_config.yml` file instead. Are there other methods I could try, or should I leave this issue? @mmcky

mmcky:
thanks @xuanguang-li we will always keep the css in `quantecon-book-theme` but thanks for giving that a go. 

The `collapse-20` changes the code cell to be only 20 lines wide and then adds a scroll, but I agree with you it makes it very hard to read the code. 

- [ ] review style of `collapse-##` in the `quantecon-book-theme` as this has changed (perhaps downstream in `sphinx_book_theme`. It should expand out and collapse (as we had before). 

@xuanguang-li can you make a new issue based on this issue for tracking? I would suggest using `collapse-40` for now perhaps.

xuanguang-li:
Thanks @mmcky. I have created a new issue QuantEcon/quantecon-book-theme#299 describing this problem.

xuanguang-li:
This PR is ready for review. @mmcky @HumphreyYang

lectures/von_neumann_model.md
```diff
@@ -107,13 +109,13 @@ class Neumann(object):
         assert (self.A >= 0).all() and (self.B >= 0).all(), 'The input and \
               output matrices must have only non-negative entries!'
 
-        # (1) Check whether Assumption I is satisfied:
+        # (1) Check whether Assumption 1 is satisfied:
```
mmcky: @xuanguang-li could you revert this to roman numerals? I think we have kept the code the same, so these should also be roman numeral. Is that right?  Thanks.

mmcky (approved):
thanks @xuanguang-li. Just one minor pickup. 

@HumphreyYang I think this is in pretty good shape.

lectures/von_neumann_model.md
```diff
@@ -546,9 +552,11 @@ Two problems are connected by a remarkable dual
 relationship between  technological and valuation characteristics of
 the economy:
 
-**Definition:** The *technological expansion problem* (TEP) for the economy
```
HumphreyYang, line 549: Here I think we want to close the definition environment after the equation.

image

lectures/von_neumann_model.md
```diff
@@ -129,8 +131,8 @@ class Neumann(object):
           - number of goods          : {n}
           - number of activities     : {m}
 
-        Assumptions:
-          - AI:  every column of B has a positive entry    : {AI}
+        Assumptions://
+          - AI:  every column of B has a positive entry    : {AI}
```
HumphreyYang:
```suggestion
        Assumptions:
          - AI:  every column of B has a positive entry    : {AI} 
```

HumphreyYang (commented):
Many thanks, @xuanguang-li! Great PR!

Just two very minor comments below.

Another small comment is that I think this lecture is not yet style-sheet compliant, as some paragraphs contain more than one sentence : )

mmcky:
thanks @HumphreyYang. 

@xuanguang-li it would be great if you could read through https://manual.quantecon.org/styleguide/writing.html and see if you make some improvements re: style guide compatibility.

xuanguang-li:
Thanks for the comments @mmcky @HumphreyYang. 

I will correct some typos and check the style again with the guidance.

mmcky:
thanks for your work on this @xuanguang-li -- this is getting close.

xuanguang-li:
Hi @mmcky @HumphreyYang,

I have divided the paragraphs into one sentence and used bold font to emphasize the definitions. 

Do you have any suggestions or further improvements I could make?

HumphreyYang:
Many thanks @xuanguang-li and @mmcky! These changes are great improvements!

I just noticed that we might want to change `^T` to `^\top` according to the style sheet, and some minor typos. 

I have collected them in one push above.

Once it builds I will merge if @mmcky agrees

xuanguang-li:
Thanks for your help, @HumphreyYang. I will keep this point in mind next time.

HumphreyYang:
Hi @xuanguang-li, I think from previous conversation, we want `collapse-40` instead of `collapse-20` to be there. Please let me know if I missed anything : )

xuanguang-li:
Yes, generally it is. @HumphreyYang 

But in issue #478, John says he wants to show all the code of `class Nuemann`.

> Only part of `class Neumann(object):` is showing --- show all

HumphreyYang:
> Yes, generally it is. @HumphreyYang
> 
> But in issue #478, John says he wants to show all the code of `class Nuemann`.
> 
> > Only part of `class Neumann(object):` is showing --- show all

Many thanks @xuanguang-li, roger that!

mmcky:
@xuanguang-li as Humphrey suggests, let's show all for now (i.e. remove the collapse tag). I have an issue to look into the old behaviour where we could expand and show-all rather than scroll.

xuanguang-li:
Got it, @mmcky! I'll drop the `collapse-##` tag when I meet it next time.

HumphreyYang:
Many thanks @xuanguang-li and @mmcky! Merging it now!

PR #569, opened 2025-08-22
[ar1_bayes] Removed PyMC dependency and style-sheet compliance

HumphreyYang:
Key changes:
- **Removed PyMC**: Removed all PyMC imports, installations, and model implementations 
- **Updated variable naming**: Used Greek letters (ρ, σ) instead of `rho, sigma` in NumPyro code
- **Code cleanup**: 
  - Cleaned up NumPyro implementation while preserving discussions
  - Fixed minor typos ("kernal" → "kernel")

### DIFF
--- a/lectures/ar1_bayes.md
+++ b/lectures/ar1_bayes.md
@@ -131,15 +131,14 @@ To illustrate the issue, we'll begin by choosing an initial $y_0$ that is far ou
 
 ```{code-cell} ipython3
 def ar1_simulate(ρ, σ, y0, T, key):
-    ε = random.normal(key, shape=(T,)) * σ
-
-    def scan_fn(y_prev, ε_t):
-        y_t = ρ * y_prev + ε_t
-        return y_t, y_t
-
-    _, y = lax.scan(scan_fn, y0, ε)
-
-    return y
+  y = [y0]
+  ε = random.normal(key, (T,)) * σ
+  
+  for t in range(T):
+    y_next = ρ * y[-1] + ε[t]
+    y.append(y_next)
+  
+  return jnp.array(y)
 
 σ = 1.0
 ρ = 0.5
--- a/lectures/ar1_bayes.md
+++ b/lectures/ar1_bayes.md
@@ -37,14 +37,11 @@ import matplotlib.pyplot as plt
 
 This lecture uses Bayesian methods offered by [`numpyro`](https://num.pyro.ai/en/stable/) to make statistical inferences about two parameters of a univariate first-order autoregression.
 
-
 The model is a good laboratory for illustrating the
 consequences of alternative ways of modeling the distribution of the initial  $y_0$:
 
 - As a fixed number
-
 - As a random variable drawn from the stationary distribution of the $\{y_t\}$ stochastic process
-
 
 The first component of the statistical model is
 
@@ -61,8 +58,6 @@ The second component of the statistical model is
 $$
 y_0 \sim {\mathcal{N}}(\mu_0, \sigma_0^2)
 $$ (eq:themodel_2)
-
-
 
 Consider a sample $\{y_t\}_{t=0}^T$ governed by this statistical model.
 
@@ -88,7 +83,6 @@ We want to study how inferences about the unknown parameters $(\rho, \sigma_x)$ 
 Below, we study two widely used alternative assumptions:
 
 -  $(\mu_0,\sigma_0) = (y_0, 0)$ which means that $y_0$ is drawn from the distribution ${\mathcal N}(y_0, 0)$; in effect, we are *conditioning on an observed initial value*.
-
 -  $\mu_0,\sigma_0$ are functions of $\rho, \sigma_x$ because $y_0$ is drawn from the stationary distribution implied by $\rho, \sigma_x$.
 
 
@@ -125,17 +119,13 @@ Basically, when $y_0$ happens to be in the tail of the stationary distribution a
 
 An example below shows how not conditioning on $y_0$ adversely shifts the posterior probability distribution of $\rho$ toward larger values.
 
-
 We begin by solving a *direct problem* that simulates an AR(1) process.
 
 How we select the initial value $y_0$ matters:
 
-   * If we think $y_0$ is drawn from the stationary distribution ${\mathcal N}(0, \frac{\sigma_x^{2}}{1-\rho^2})$, then it is a good idea to use this distribution as $f(y_0)$. 
-  
-       -  Why? Because $y_0$ contains information about $\rho, \sigma_x$.
-
-   * If we suspect that $y_0$ is far in the tail of the stationary distribution -- so that variation in early observations in the sample has a significant *transient component* -- it is better to condition on $y_0$ by setting $f(y_0) = 1$.
-
+* If we think $y_0$ is drawn from the stationary distribution ${\mathcal N}(0, \frac{\sigma_x^{2}}{1-\rho^2})$, then it is a good idea to use this distribution as $f(y_0)$. 
+  -  Why? Because $y_0$ contains information about $\rho, \sigma_x$.
+* If we suspect that $y_0$ is far in the tail of the stationary distribution -- so that variation in early observations in the sample has a significant *transient component* -- it is better to condition on $y_0$ by setting $f(y_0) = 1$.
 
 To illustrate the issue, we'll begin by choosing an initial $y_0$ that is far out in the tail of the stationary distribution.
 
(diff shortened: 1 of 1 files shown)


lectures/ar1_bayes.md
```diff
@@ -355,11 +224,21 @@ mcmc.run(rng_key=random.PRNGKey(1), data=y)
 plot_posterior(mcmc.get_samples())
 ```
 
+Evidently, the posteriors aren't centered on the true values of $.5, 1$ that we used to generate the data.
+
+This is a symptom of the classic **Hurwicz bias** for first order autoregressive processes (see Leonid Hurwicz {cite}`hurwicz1950least`.)
```
mmcky:
```suggestion
This is a symptom of the classic **Hurwicz bias** for first order autoregressive processes (see  {cite}`hurwicz1950least`.)
```

mmcky (commented):
Thanks @HumphreyYang -- just one minor comment re: citation. Should I organise @thomassargent30 to review?

HumphreyYang:
> Thanks @HumphreyYang -- just one minor comment re: citation. Should I organise @thomassargent30 to review?

Thanks Matt, I think it's fine. This is a minor change : )

mmcky (approved):

mmcky:
thanks @HumphreyYang are you happy for me to merge this?

jstac:
@mmcky Where do I find the latest deployment? 

I can see above "🚀 Deployed on https://68bb28185ee35015bb2cce79--nostalgic-wright-5fa355.netlify.app/"

I'm guessing this is out of date because it's not compliant with the syle manual (e.g., sentences running together instead of being followed by line breaks).

I find this PR a bit hard to jump into quickly because I'm unsure of the deployment and the title of the lecture is missing.

HumphreyYang:
> @mmcky Where do I find the latest deployment?
> 

Many thanks @jstac, please see the latest deployment of the lecture **Posterior Distributions for AR(1) Parameters**:
https://68bd39bc9ca87aaef6818d45--nostalgic-wright-5fa355.netlify.app/ar1_bayes

We will include the lecture title and the latest deployment of the lecture next time we ping you for review.

Many thanks in advance!

(I pushed some more updates since I found some more paragraphs with multiple sentences)

mmcky:
@HumphreyYang doing some testing of the `style-guide` compliance using copilot. We can incorporate the useful components from #593 once it is completed.

mmcky:
@copilot I have attached our style-guide `quantecon_review_instructions.md`. 

[quantecon_review_instructions.md](https://github.com/user-attachments/files/22199445/quantecon_review_instructions.md)

Can you please review this lecture `ar1_bayes` and make sure it is style guide compliant. 

Are you able to do a full review of this lecture and not just a review of the changes.

mmcky:
> @copilot I have attached our style-guide `quantecon_review_instructions.md`.
> 
> [quantecon_review_instructions.md](https://github.com/user-attachments/files/22199445/quantecon_review_instructions.md)
> 
> Can you please review this lecture `ar1_bayes` and make sure it is style guide compliant.
> 
> Are you able to do a full review of this lecture and not just a review of the changes.

@HumphreyYang this sort of triggering doesn't work. I think it will only work when a PR is authored initially by `copilot`. I will do a full pass of this lecture on my local machine re: style compliance.

HumphreyYang:
Many thanks @mmcky, this PR should be compliant since I have run the same instruction on my end, but please let me know if your run spots anything new!

mmcky:
thanks @HumphreyYang -- you're right. I'll open an issue to figure out how we can improve the `netlify` deployment links. I find it interesting it doesn't post the latest one after each build.

HumphreyYang:
> thanks @HumphreyYang -- you're right. I'll open an issue to figure out how we can improve the `netlify` deployment links. I find it interesting it doesn't post the latest one after each build.

Many thanks @mmcky! It's interesting that it posts deployment this time : )

mmcky:
@HumphreyYang it looks like `netlify` deployments disappear! Maybe that is part of the issue we are having.

lectures/ar1_bayes.md
```diff
@@ -103,68 +82,71 @@ We want to study how inferences about the unknown parameters $(\rho, \sigma_x)$
...
 
+* If we think $y_0$ is drawn from the stationary distribution ${\mathcal N}(0, \frac{\sigma_x^{2}}{1-\rho^2})$, then it is a good idea to use this distribution as $f(y_0)$. 
+  -  Why? Because $y_0$ contains information about $\rho, \sigma_x$.
+* If we suspect that $y_0$ is far in the tail of the stationary distribution -- so that variation in early observations in the sample has a significant *transient component* -- it is better to condition on $y_0$ by setting $f(y_0) = 1$.
 
-To illustrate the issue, we'll begin by choosing an initial $y_0$ that is far out in a tail of the stationary distribution.
+To illustrate the issue, we'll begin by choosing an initial $y_0$ that is far out in the tail of the stationary distribution.
 
 ```{code-cell} ipython3
+def ar1_simulate(ρ, σ, y0, T, key):
```
jstac, line 133: @HumphreyYang 

Many thanks for all your great work in this PR.  One comment:

IMO, the QE rule should be always use the simplest solution unless we can get a significant performance gain and that gain matters.  But that rule, we should be appending to an ordinary Python list rather than using `lax.scan` inside this function.

HumphreyYang: Many thanks @jstac. Roger that, I will update this.

I used this approach because I found scan syntax is more concise than the for loop which needs initialization and updating : )

But I agree with the simple-solution principle!

jstac: > I found scan syntax is more concise than the for loop which needs initialization and updating : )

I agree with you @HumphreyYang and I can see the elegance, but we need to detune our code and go for elementary solutions when trying to communicate with humans who live out in the wild :-)

Please also enforce this principle with the junior RAs.

HumphreyYang: Thanks @jstac, roger that!

jstac:
I find this code quite confusing:

```python
def AR1_model(data):
    # Set prior
    ρ = numpyro.sample('ρ', dist.Uniform(low=-1., high=1.))
    σ = numpyro.sample('σ', dist.HalfNormal(scale=jnp.sqrt(10)))

    # Expected value of y in the next period (ρ * y)
    yhat = ρ * data[:-1]

    # Likelihood of the actual realization
    numpyro.sample('y_obs', 
                dist.Normal(loc=yhat, scale=σ), obs=data[1:])
```

I think I've gotten used to JAX's functional style... Nothing is being returned here.  How does it work?

Maybe I'm the only one who finds this confusing.  Or maybe we need some discussion around the code to explain how it works...

HumphreyYang:
> I think I've gotten used to JAX's functional style... Nothing is being returned here. How does it work?

Hi @jstac, I think that concern is totally fair. I was confused when I first saw this syntax. NumPyro uses a functional core with an imperative shell: the function here is just a specification (like writing specification down in `sympy`), and a stack of effect handlers (dynamic, thread-local) reinterprets each `numpyro.sample(...)` during execution.

When we call `mcmc.run(...)`, the kernel (e.g., NUTS) builds a pure log-density from the model via these handlers and then does the sampling. So the handlers don’t do sampling per sa; they construct the potential function the sampler needs.

It can be a bit much for readers who just want posterior draws, so I can add a few simple notes next to the first NumPyro code cells in the previous lecture to demystify this a little.

If that sounds good, I’ll open a separate issue and PR.

jstac:
Thanks @HumphreyYang , that would be great.

I'm starting to really appreciate the explicitness and clean logic of JAX.

jstac:

**Netlify preview:** https://pr-569--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 181 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

### ⭐ Directly relevant to this PR
This PR edits **`ar1_bayes.md`** — one of the two lectures that tripped the arviz-plots `figsize` bug. It is **already fixed upstream in arviz 1.2.0** (verified: `az.plot_trace` on a real pymc AR1 posterior renders on the latest stack). So there is *nothing to fix in the arviz code here* — focus on updating the branch to `main` and rebuilding.

Update this branch to `main` (merge or rebase — pulls in #939 plus ~181 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-569--sunny-cactus-210e3e.netlify.app/ar1_bayes.html
```

---
*This PR touches: `ar1_bayes.md`. Last CI build: `success@2025-09-19`. Branch: 181 commits behind `main` as of 2026-07-08.*

jstac:
Many thanks @HumphreyYang .  I worked on this lecture in a separate PR but took most of your changes.  I ended up deciding to keep PyMC since it's useful to verify the result from two methods.  For more sophisticated problems we'll favor Numpyro / JAX unless we have good reasons not to.  Closing.

PR #584, opened 2025-08-31
[ar1_turningpts] Removed PyMC and changed the style 

xuanguang-li:
Key changes:

- **Removed PyMC**: Used `Numpyro` to compute posteriors instead.
- **Refactored class and functions**: Defined an `AR1` class to pack parameters, separated plotting functions, and packed path statistics computation functions.
- **Removed `for` loops**: Used `lax.scan` in almost all loops except for plotting functions and the `Numpyro` model.
- **Removed `numpy.random`**: Used `jax.random` for simulation and fixed the random seed with `PRNGKey(0)`.
- **Fixed typos and styling**: Followed the style guide and PEP8.

This improved the efficiency of the main function from 3m to 2.7s on my computer!


lectures/ar1_turningpts.md
```diff
@@ -91,12 +99,16 @@ We form this predictive distribution by integrating {eq}`ar1-tp-eq3` with respec
...
+&= \int f(y_{t+j} | y^t, \rho, \sigma) \pi_t(\rho,\sigma | y^t ) d \rho d \sigma\\
+&= \int f(y_{t+j} | y_t, \rho, \sigma) \pi_t(\rho,\sigma | y^t ) d \rho d \sigma
+\end{aligned}
 $$ (ar1-tp-eq4)
 
-Predictive distribution {eq}`ar1-tp-eq3` assumes that parameters $(\rho,\sigma)$ are known.
+Predictive distribution {eq}`ar1-tp-eq3` assumes that parameters $(\rho,\sigma)$ are known. 
 
-Predictive distribution {eq}`ar1-tp-eq4` assumes that parameters $(\rho,\sigma)$ are uncertain, but have known probability distribution $\pi_t(\rho,\sigma | y^t )$.
+Predictive distribution {eq}`ar1-tp-eq4` assumes that parameters $(\rho,\sigma)$ are uncertain, but have known probability distribution $\pi_t(\rho,\sigma | y^t )$. Notice the second equality follows that $\{y_t\}$ is a AR(1) process when $(\rho, \sigma)$ are given.
```
mmcky:
```suggestion
Predictive distribution {eq}`ar1-tp-eq4` assumes that parameters $(\rho,\sigma)$ are uncertain, but have known probability distribution $\pi_t(\rho,\sigma | y^t )$. 

(Notice the second equality follows that $\{y_t\}$ is a AR(1) process when $(\rho, \sigma)$ are given.)  
```

mmcky:
@xuanguang-li this is a big translation effort. Nice work! @HumphreyYang would you have time to review (no rush).

mmcky:
@xuanguang-li sorry I just saw you have applied the `in-work` label. Let me know when you're ready for this to be reviewed.

xuanguang-li:
Thanks @mmcky. I will remove the `in-work` label after checking the typos in one or two days.

lectures/ar1_turningpts.md
```diff
@@ -123,62 +135,184 @@ we'll plot $.9$ and $.95$ coverage intervals using conditional distribution
...
+    T0 : int, optional
+        Length of the initial observed path (default is 100).
+    T1 : int, optional
+        Length of the future path to simulate (default is 100).
+    """
+    rho: float
+    sigma: float
+    y0: float
+    T0: int=100
+    T1: int=100
```
HumphreyYang: I think we would prefer to set default parameters in the parameter factory.

lectures/ar1_turningpts.md
```diff
@@ -123,62 +135,184 @@ we'll plot $.9$ and $.95$ coverage intervals using conditional distribution
...
+        AR1 named tuple containing parameters (rho, sigma, y0, T0, T1).
+    key : jax.random.PRNGKey
+        JAX random key for generating random noise.
+
+    Returns
+    -------
+    initial_path : jax.numpy.ndarray
+        Simulated path of the AR(1) process and the initial y0.
+    """
+    rho, sigma, y0, T0 = ar1.rho, ar1.sigma, ar1.y0, ar1.T0
```
HumphreyYang: Replace

`sigma` -> σ 
`rho` -> ρ

in the code.

Just be careful that unicode might not work in `numpyro.sample` like:

```
rho = numpyro.sample('rho', dist.Uniform(-1, 1)) 
sigma = numpyro.sample('sigma', dist.HalfNormal(jnp.sqrt(10)))
```

For those, we can write

```
ρ = numpyro.sample('rho', dist.Uniform(-1, 1)) 
σ = numpyro.sample('sigma', dist.HalfNormal(jnp.sqrt(10)))
```

lectures/ar1_turningpts.md
```diff
@@ -123,62 +135,184 @@ we'll plot $.9$ and $.95$ coverage intervals using conditional distribution
...
+    # Draw epsilons
+    eps = sigma * random.normal(key, (T0,))
+    # Set step function
+    def ar1_step(y_prev, t_rho_eps):
+        rho, eps_t = t_rho_eps
+        y_t = rho * y_prev + eps_t
```
HumphreyYang: Replace

`eps` -> ε

in the code.

xuanguang-li:
Hi @mmcky and @HumphreyYang, this PR is ready for review.

lectures/ar1_turningpts.md
```diff
@@ -233,240 +367,251 @@ $$
...
+        fig, ax = plt.subplots(1, 2, figsize=(12, 5))
+        sns.histplot(
+            post_sample['rho'], kde=True, stat="density", bins=bins, ax=ax[0]
+        )
+        ax[0].set_xlabel(r"$\rho$")
+        sns.histplot(
+            post_sample['sigma'], kde=True, stat="density", bins=bins, ax=ax[1]
+        )
+        ax[1].set_xlabel(r"$\sigma$")
     return post_sample
```
HumphreyYang: I think here we are missing a trace plot?

We can also use `arviz` for `numpyro` using a very similar syntax to the `pymc` case. 

Please see code cell 4 - 5 in the documentation here for an example of using `arviz.from_numpyro()` and `az.plot_trace()`:
https://num.pyro.ai/en/0.15.1/tutorials/censoring.html

lectures/ar1_turningpts.md
```diff
@@ -13,10 +13,13 @@ kernelspec:
...
+```
+
 ```{code-cell} ipython3
 :tags: [hide-output]
 
-!pip install arviz pymc
+!pip install numpyro jax
 ```
 
 This lecture describes methods for forecasting statistics that are functions of future values of a univariate autogressive process.
```
HumphreyYang:
```suggestion
This lecture describes methods for forecasting statistics that are functions of future values of a univariate autoregressive process.  
```

lectures/ar1_turningpts.md
```diff
@@ -478,56 +623,50 @@ Now we apply we apply our  "extended" Wecker method based on  predictive densiti
 To approximate  the intergration on the right side of {eq}`ar1-tp-eq4`, we  repeatedly draw parameters from the joint posterior distribution each time we simulate a sequence of future values from model {eq}`ar1-tp-eq1`.
```
HumphreyYang:
```suggestion
To approximate  the integration on the right side of {eq}`ar1-tp-eq4`, we  repeatedly draw parameters from the joint posterior distribution each time we simulate a sequence of future values from model {eq}`ar1-tp-eq1`.
```

lectures/ar1_turningpts.md
```diff
@@ -233,240 +367,251 @@ $$
...
 
-    if np.any(z) == False:
-        return 500
-    else:
-        return np.where(z==1)[0][0] + 1
+These two kinds of definitions are equivalent because $\omega$ determins path statistics only through $\{Y_t\}$.
 
-def minimum_value(omega):
-    return min(omega[:8])
+Moreover, we ignore all equality in the definitions, as equality occurs with zero probablity for countinuous random variables.
```
HumphreyYang:
```suggestion
Moreover, we ignore all equality in the definitions, as equality occurs with zero probability for continuous random variables.
```

lectures/ar1_turningpts.md
```diff
@@ -123,62 +135,184 @@ we'll plot $.9$ and $.95$ coverage intervals using conditional distribution
...
+    y_upper_c95 = center + 1.96 * jnp.sqrt(vars)
+    y_lower_c95 = center - 1.96 * jnp.sqrt(vars)
+    # 90% CI
+    y_upper_c90 = center + 1.65 * jnp.sqrt(vars)
+    y_lower_c90 = center - 1.65 * jnp.sqrt(vars)
 
+    # Plot
+    ax.plot(jnp.arange(-T0, 1), initial_path)
+    ax.axvline(0, linestyle='--', alpha=.4, color='k', lw=1)
+    # Chooise 10 future paths to plot
```
HumphreyYang:
```suggestion
    # Choose 10 future paths to plot
```

lectures/ar1_turningpts.md
```diff
@@ -233,240 +367,251 @@ $$
...
-    sns.histplot(next_down_turn, kde=True, stat='density', ax=ax[2, 1], alpha=.8, label='True parameters')
-    ax[2, 1].set_title("Predictive distribution of time until the next negative turn", fontsize=13)
-
-fig, ax = plt.subplots(3, 2, figsize=(15,12))
-plot_Wecker(initial_path, 1000, ax)
+
+fig, ax = plt.subplots(3, 2, figsize=(15, 12))
+plot_Wecker(ar1, initial_path, ax)
 plt.show()
 ```
```
HumphreyYang, line 668: In the sentence below this line:

```
Now we apply we apply our  "extended" Wecker method -> Now we apply our  "extended" Wecker method
```

lectures/ar1_turningpts.md
```diff
@@ -233,240 +367,251 @@ $$
...
+        return carry, (rec, sev, up, down)
     
-    for i in range(n):
-        if ((omega[i] > omega[i+1]) and (omega[i+1] > omega[i+2]) and 
-            (omega[i+2] < omega[i+3]) and (omega[i+3] < omega[i+4])):
-            T[i] = 1
-        elif ((omega[i] < omega[i+1]) and (omega[i+1] < omega[i+2]) and 
-            (omega[i+2] > omega[i+3]) and (omega[i+3] > omega[i+4])):
-            T[i] = -1
+    _, (rec_seq, sev_seq, up_seq, down_seq) = lax.scan(step, None, jnp.arange(3, n-3))
```
HumphreyYang: Should we use `jnp.arange(3, n-2)` since we are only looking ahead 2 terms?

lectures/ar1_turningpts.md
```diff
@@ -233,240 +367,251 @@ $$
...
+    # Concatenate the last two elements of initial path to identify recession
+    y = jnp.concatenate([initial_path[-3:], future_path])
+    n = y.shape[0]
+    def step(carry, i):
+        # identify recession
+        rec_cond = (y[i] < y[i-1]) & (y[i-1] < y[i-2]) & (y[i-2] > y[i-3])
+        # identify severe recession
+        sev_cond = (
+            (y[i] - y[i-1] < -0.02) & (y[i-1] - y[i-2] < -0.02) & (y[i-2] > y[i-3])
+            )
```
HumphreyYang, line 531: I think the logic here is different from the old code, where we only checked for a single-period decline:

```
def severe_recession(omega):
    z = np.diff(omega)
    n = z.shape[0]
    
    sr = (z < -.02).astype(int)
    indices = np.where(sr == 1)[0]

    if len(indices) == 0:
        return T1
    else:
        return indices[0] + 1
```

But I think your definition makes sense—it needs to be classified as a recession before being classified as a severe recession.

We might also want to update the text to make it clear to the reader what criterion we are using.

lectures/ar1_turningpts.md
```diff
@@ -233,240 +367,251 @@ $$
...
-    next_up_turn, next_down_turn = np.zeros(N), np.zeros(N)
+    # Plot simulated initial and future paths
+    y_T0 = initial_path[-1]
+    future_path = AR1_simulate_future(ar1, y_T0, N=N)
+    plot_path(ar1, initial_path, future_path, ax[0, 0])
+
+    # Simulate future paths and compute statistics
+    def step(carry, n):
+        future_temp = future_path[n, :]
+        (next_reces, severe_rec, min_vals, next_up_turn, next_down_turn
```
HumphreyYang: Is `min_vals` the same as `min_val_8q`? It would be nice if we can unify the variable names. I think `min_vals` is more concise!

HumphreyYang (changes requested):
This is a really nice PR — a lot of improvements that make everything clean and concise.

Many thanks, @xuanguang-li!

I really like how the recession checks are improved, saving a lot of lines of code. Well done!

Of course, for a PR that has this amount of substantial and helpful changes, we will run over a few iterations!

**General comments:**

1. Check grammar throughout the lecture (this can be done using Claude code).
2. Leave some space between a comment and the code above it, e.g.,

   ```python
    # Split key to generate different paths
    subkeys = random.split(key, num=N)

    # Simulate N paths
    future_path = jax.vmap(single_path_scan, in_axes=(None, 0))(y_T0, subkeys)
   ```
    instead of 
    ```python
    # Split key to generate different paths
    subkeys = random.split(key, num=N)
    # Simulate N paths
    future_path = jax.vmap(single_path_scan, in_axes=(None, 0))(y_T0, subkeys)
    ```
    This way, the code doesn’t look too dense.

3. For larger code blocks, it would be nice to break them up with some comments in between. For instance, between the code that defines `AR1_simulate_past` and `AR1_simulate_future`, you could add a transition sentence like:

   > “Now we define the simulation function that generates a realization of the AR(1) process for \$T1\$ periods.”

   This will help make the big code blocks easier to follow.

xuanguang-li:
Thanks, @HumphreyYang, for your helpful suggestions!

lectures/ar1_turningpts.md
xuanguang-li: Yes, that is what I thought - a severe recession should first be a recession. I added one sentence after the definition of the recession to briefly introduce this concept:

> Moreover, the statistic *time until the next severe recession* is defined in a similar way, except the decline between periods is greater than $0.02$.

I will refer to Wecker's paper to see whether my definition is consistent with his.

xuanguang-li:
Hi @HumphreyYang,

I tried `arviz` to plot the trace, and to realize the same trace as before, I set the `num-chains=4` in the `mcmc.run`.

However, it results in the following error, as it is not simulated by 4 devices.

> UserWarning: There are not enough devices to run parallel chains: expected 4 but got 1. Chains will be drawn sequentially. If you are running MCMC in CPU, consider using `numpyro.set_host_device_count(4)` at the beginning of your program.

Do you think we should recover the `num-chains=1` to avoid this error or leave the error there?

HumphreyYang:
Many thanks, @xuanguang-li!

> Do you think we should set `num-chains=1` to avoid this error, or should we just leave the error as is?

In `numpyro`, the number of chains is parallelized across the available processors. So, to run four chains, you can simply set `numpyro.set_host_device_count(4)`.

xuanguang-li:
Thanks, @HumphreyYang. 

I have set `numpyro.set_host_device(4)` at the `import` cell, but it did not resolve this error, as shown in the Github deploy. Should it be set inside the `MCMC.run`?

HumphreyYang:
Hi @xuanguang-li, nice observation. This is because we need to set host device before we import JAX:

```
import matplotlib.pyplot as plt
import seaborn as sns
from typing import NamedTuple

# numpyro
import numpyro
import numpyro.distributions as dist
from numpyro.infer import MCMC, NUTS
numpyro.set_host_device_count(4)

# jax
import jax
import jax.random as random
import jax.numpy as jnp
from jax import lax

# arviz
import arviz as az

sns.set_style('white')
colors = sns.color_palette()
key = random.PRNGKey(0)
```

This should fix the warning.

(See the note in https://num.pyro.ai/en/stable/utilities.html#set-host-device-count).

xuanguang-li:
Hi @HumphreyYang, I’m sorry but this method didn’t eliminate the error report in the GitHub deployment.  

I printed `jax.local_device_count()` at the bottom of the `import` cell, and it still shows the device as 1. However, this method did work on my local machine.  

Do you have any idea what might be happening?

HumphreyYang:
Many thanks @xuanguang-li, Yeah I see why. That's because the series is running on a GPU server with just one GPU. I was testing on my local machine with multicore CPU. 

For GPUs, `nupyro` parallization strategy is `chain_method="vectorized"`. I just pushed a commit and it should fix the warning.

Also, I think the latest `numpyro` version can take a unicode string as variable name now (i.e. in `numpyro.sample`)

xuanguang-li:
Thanks @HumphreyYang! This solved the problem perfectly.  

> Also, I think the latest `numpyro` version can take a unicode string as a variable name now (i.e. in `numpyro.sample`)  

I’ll try using Greek letters in `numpyro.sample`. Thanks for letting me know!

xuanguang-li:
Hi @mmcky @HumphreyYang, I’ve implemented the requested changes. Could you take a look when you have time?  

Regarding **severe recession**, I couldn’t find a clear definition in Wecker (1979), and the NBER website’s definition seems vague:  

> In our interpretation of this definition, we treat the three criteria — depth, diffusion, and duration — as somewhat interchangeable.

https://www.nber.org/research/business-cycle-dating

jstac:
We don't need an instantiation function here.  Defaults can be stored in the class. 

```python
class AR1(NamedTuple):
    """
    Represents a univariate first-order autoregressive (AR(1)) process.

    Parameters
    ----------
    ρ : float
        Autoregressive coefficient, must satisfy |ρ| < 1 for stationarity.
    σ : float
        Standard deviation of the error term.
    y0 : float
        Initial value of the process at time t=0.
    T0 : int, optional
        Length of the initial observed path (default is 100).
    T1 : int, optional
        Length of the future path to simulate (default is 100).
    """
    ρ: float
    σ: float
    y0: float
    T0: int
    T1: int


def make_ar1(ρ: float, σ: float, y0: float, T0: int = 100, T1: int = 100):
    """
    Factory function to create an AR1 instance with default values for T0 and T1.
    
    Returns
    -------
    AR1
        AR1 named tuple containing the specified parameters.
    """
    return AR1(ρ=ρ, σ=σ, y0=y0, T0=T0, T1=T1)
```

jstac:
Link needs fixing?

image

jstac:
I'm not in favor of dropping down into `lax` in functions like this -- plotting code where speed is irrelevant.

I recommend using `at[x].set(y)` or a appending to an ordinary Python list.  What do you think @mmcky @HumphreyYang ?



```python
def plot_Wecker(ar1: AR1, initial_path, ax, N=1000):
    """
    Plot the predictive distributions from "pure" Wecker's method.

    Parameters
    ----------
    ar1 : AR1
        An AR1 named tuple containing the process parameters (ρ, σ, T0, T1).
    initial_path : array-like
        The initial observed path of the AR(1) process.
    N : int
        Number of future sample paths to simulate for predictive distributions.
    """
    # Plot simulated initial and future paths
    y_T0 = initial_path[-1]
    future_path = AR1_simulate_future(ar1, y_T0, N=N)
    plot_path(ar1, initial_path, future_path, ax[0, 0])

    # Simulate future paths and compute statistics
    def step(carry, n):
        future_temp = future_path[n, :]
        (next_reces, severe_rec, min_val_8q, next_up_turn, next_down_turn
         ) = compute_path_statistics(initial_path, future_temp)
    
        return carry, (next_reces, severe_rec, min_val_8q, next_up_turn, next_down_turn)
    
    _, (next_reces, severe_rec, min_val_8q, next_up_turn, next_down_turn
        ) = lax.scan(step, None, jnp.arange(N))
    
    # Plot path statistics
    plot_path_stats(next_reces, severe_rec, min_val_8q, 
                    next_up_turn, next_down_turn, ax)


fig, ax = plt.subplots(3, 2, figsize=(15, 12))
plot_Wecker(ar1, initial_path, ax)
plt.show()
```

xuanguang-li:
Hi @jstac, thanks for your comment.

> Link needs fixing?

Do you mean we should write the link as 
`[Permanent Income II: LQ Techniques](https://python.quantecon.org/perm_income_cons)`?

jstac:
> Hi @jstac, thanks for your comment.
> 
> > Link needs fixing?
> 
> Do you mean we should write the link as `[Permanent Income II: LQ Techniques](https://python.quantecon.org/perm_income_cons)`?

Thanks @xuanguang-li .  I can't remember but it should be in the manual -- or @mmcky will know.

xuanguang-li:
Thanks for your reminding @jstac. 

I found it in the manual *Referencing a lecture in a different lecture series*.

mmcky:
thanks @xuanguang-li -- glad you found it https://manual.quantecon.org/styleguide/doclinks.html#referencing-a-lecture-in-a-different-lecture-series -- let me know if we need to configure `intersphinx`. Happy to help with that.

xuanguang-li:
Hi @jstac, I have implemented the review comments in the latest commit.

Thanks, @mmcky. I realized this link points to a lecture in the same series, so we can reference it directly.

However, a warning is raised when using the standard Markdown link `[](perm_income_cons)` because `perm_income_cons` also refers to a section within a lecture. I checked the *MyST syntax cheat sheet* and found that using ``{doc}`perm_income_cons`` resolves this issue. Maybe we could add this method to *Referencing another lecture* section of the manual to avoid this problem in the future.

jstac:
Thanks @xuanguang-li .  We appreciate you thinking carefully about this and correctly diagnosing the problem.

But we always want consistency @xuanguang-li , not special cases.  The best action here is to change the name of the subsection and anything that refers to it --- and then use the standard method from the manual for the link to the lecture.

Such a section name should never have been created -- since it interferes with links across lectures -- and this is a good opportunity to fix it.

xuanguang-li:
Thanks @jstac. Sorry, I didn't look into the problem explicitly. You are right; no subsection used this label.

According to the log, the issue seems to arise because `perm_income_cons` also appears in the *Dynamic Linear Economics* series:

```WARNING: Multiple matches found for 'perm_income_cons': dle:std:label:perm_income_cons, dle:std:doc:perm_income_cons, intermediate:std:doc:perm_income_cons, ... [myst.iref_ambiguous] ```

I agree that we should avoid special cases, but I think we are not making an exception because the `{doc}` reference is used frequently for within-series reference.

**Comment by jstac** [comment]
Thanks for further thoughts @xuanguang-li .  They are appreciated.

@mmcky -- perhaps you could add your input?

**Comment by jstac** [comment]

**Netlify preview:** https://pr-584--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 181 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

### ⭐ Directly relevant to this PR
This PR edits **`ar1_turningpts.md`** — one of the two lectures that tripped the arviz-plots `figsize` bug. It is **already fixed upstream in arviz 1.2.0** (verified: `az.plot_trace` on a real pymc AR1 posterior renders on the latest stack). So there is *nothing to fix in the arviz code here* — focus on updating the branch to `main` and rebuilding.

Update this branch to `main` (merge or rebase — pulls in #939 plus ~181 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-584--sunny-cactus-210e3e.netlify.app/ar1_turningpts.html
```

---
*This PR touches: `ar1_turningpts.md`. Last CI build: `success@2025-09-23`. Branch: 181 commits behind `main` as of 2026-07-08.*

jstac:
Pushed fixes on top of this branch (merged latest `main`) to get it building. GPU-verified end-to-end on an RTX 4080 (NUTS + JAX, ~110s, all 5 figures render, 0 errors).

**Correctness / build blockers**
- Fixed `NamedTuplep` import typo that broke the whole build (the cause of the failing CI).
- Vectorized the NumPyro likelihood — it was a per-observation Python loop creating one sample site per datum; now a single vectorized `y_obs` site, matching `ar1_bayes.md`.
- `plot_extended_Wecker` was plotting an undefined `future_path` (silently falling back to the true-parameter global); it now plots the posterior-draw paths.

**arviz 1.x**
- `az.plot_trace(data=…, compact=…, lines=…, backend_kwargs=…)` is gone in arviz 1.2.0 → switched to `az.plot_trace_dist(...)`; updated prose for the new left-column densities / right-column traces layout.

**Style / polish**
- Added a `create_ar1()` factory with validation; `@jax.jit` on `compute_path_statistics`; `lw=2` on line charts; tidied captions; reverted the jupytext header to `.myst`.
- Restored the *true parameters* vs *sampling from posterior* legend on the comparison figure.

Note: the NumPy→JAX RNG switch changes every figure's exact values (expected per the migration guide); shapes/scales and the narrative still hold.

jstac:
Many thanks @xuanguang-li for your hard work on this PR!  I've asked Claude to make some minor edits.  I'll merge when it's green.

xuanguang-li:
Thanks for your review, @jstac! The commit helped to fix many typos and style issues.

PR #545, opened 2025-08-15
[bayes_nonconj] Update lecture

kp992:
Updates the lecture with:
- Removed pyro and torch code
- Uses only numpyro backed by JAX.
- Typos and Code styling.

Fixes #524

PR #586, opened 2025-09-01
[bayes_nonconj] Test compile

xuanguang-li:
Errors occurred in the original PR #545 for the file that was successfully compiled before. This PR is used for testing compilation.


mmcky:
Thanks @xuanguang-li -- I will close this once we get #545 building again (due to the change in software including `python=3.13`)

xuanguang-li:
The error has been solved after updating to the latest `main`.

PR #589, opened 2025-09-02
[lake_model] Update lecture to JAX and check compliance to style sheet

HumphreyYang:
This PR updates `lake_model` lecture to JAX and check the compliance to style sheet. 

This reduces a lot of code in the lecture!


lectures/lake_model.md
```diff
@@ -67,14 +76,14 @@ These concepts will help us build an equilibrium model of ex-ante homogeneous wo
...
-import numpy as np
-from quantecon import MarkovChain
-from scipy.stats import norm
-from scipy.optimize import brentq
+import jax
+import jax.numpy as jnp
+from typing import NamedTuple
 from quantecon.distributions import BetaBinomial
-from numba import jit
+from functools import partial
```
mmcky, line 77: @HumphreyYang I have been thinking about the use of `partial`. I think we should add a `note` or `tip` box to explain what it is doing.

lectures/lake_model.md
```diff
@@ -209,132 +218,101 @@ This follows from the fact that the columns of $\hat A$ sum to 1.
...
-    α : scalar
-        The dismissal rate for currently employed workers
-    b : scalar
-        Entry rate into the labor force
-    d : scalar
-        Exit rate from the labor force
 
+@jax.jit
+def rate_steady_state(model, tol=1e-6):
+    r"""
```
mmcky: @HumphreyYang this `r` is required because of the `math` role right? I might update the manual.

lectures/lake_model.md
```diff
@@ -635,112 +623,99 @@ Following {cite}`davis2006flow`, we set $\alpha$, the hazard rate of leaving emp
...
 
-The second piece of code is used to complete the reservation wage:
+The second piece of code is used to compute the reservation wage:
 
-```{code-cell} python3
----
-tags: [output_scroll]
----
+```{code-cell} ipython3
+:tags: [output_scroll]
```
mmcky: I think we should move to `collapse-40` once https://github.com/QuantEcon/quantecon-book-theme/pull/300 is ready.

mmcky: @HumphreyYang let's leave this for now and we can do a bulk search and conversion once the theme is improved.

lectures/lake_model.md
```diff
@@ -209,132 +218,101 @@ This follows from the fact that the columns of $\hat A$ sum to 1.
...
+    x = x / x.sum()
+    return x
 
-As explained, if we create an instance and update it by `lm = LakeModel(α=0.03)`,
-derived objects like $A$ will also change.
 
-```{code-cell} python3
-lm = LakeModel()
-lm.α
+@partial(jax.jit, static_argnums=(2,))
```
mmcky:
```suggestion
@partial(jax.jit, static_argnames=['T'])
```

lectures/lake_model.md
mmcky: matplotlib issue

lectures/lake_model.md
HumphreyYang: Hi @mmcky,

I will leave the explanation to the JAX lecture as I think it is a common pattern in JAX.

jstac:
In this situation we don't need the separate create function because we can store the defaults in the class itself and instantiate it directly.

We only need a separate create function if we have to use default parameter values to build intermediate objects and then store them in the class.  (Maybe this should be part of the style file @mmcky @HumphreyYang ?)

```python
class LakeModel(NamedTuple):
    """
    Parameters for the lake model
    """
    λ: float
    α: float
    b: float
    d: float


def create_lake_model(λ=0.283, α=0.013, b=0.0124, d=0.00822):
    """Create a LakeModel with default parameters."""
    return LakeModel(λ=λ, α=α, b=b, d=d)

```

jstac:
I'm not sure of the right answer here but does anyone feel like this is overkill?    We're only iterating a 2x2 matrix multiplication 50 times.  Do we need to dip into JAX's low level functions?

I realize there's no great JAX alternative and I can see this is elegant.  But maybe we should just append to a Python list, or use the `at[i].set(b)` syntax.  Thoughts?  cc @HumphreyYang @mmcky @cc7768

```python
@partial(jax.jit, static_argnames=['T'])
def simulate_stock_path(model, X0, T):
    """
    Simulates the sequence of employment and unemployment stocks.
    """
    A, A_hat, g = compute_matrices(model)
    
    def update_X(X, _):
        X_new = A @ X
        return X_new, X
    
    X0 = jnp.atleast_1d(X0)
    _, X_path = jax.lax.scan(update_X, X0, jnp.arange(T))
    return X_path
```

jstac:
This, on the other hand, is super cool :-)

(Perhaps `carry` could be renamed `state`.  Then we can cut a line.  Also, should `state_new, state_new` be `state_new, state`, so that we get X_0?)

```python
@partial(jax.jit, static_argnames=['T'])
def simulate_markov_chain(P, T, init_state, key):
    """Simulate a Markov chain."""
    def step(carry, key):
        state = carry
        probs = P[state]
        state_new = jax.random.choice(key, 
                            a=jnp.arange(len(probs)), p=probs)
        return state_new, state_new
    
    keys = jax.random.split(key, T)
    _, states = jax.lax.scan(step, init_state, keys)
    return states
```

jstac:
I think it says in the style manual that matplotlib should be picking colors.  Let's let the professionals do it.

`    axes[i].hlines(x0[i], 0, T, 'r', '--')`

jstac:
Let's change this to use the standard QE style for figures, so that it doesn't stand out.  No colored background, no grid, matplotlib picks colors.

image

jstac:
From what I can see this function is never used.  (Also, for jit compiled functions, we should probably avoid situations where the return type or size depends on the *values* of the function parameters.)

```python
@jax.jit
def compute_reservation_wage(mcm, return_values=False):
    """
    Computes the reservation wage of an instance of the McCall model
    by finding the smallest w such that V(w) > U.

    If V(w) > U for all w, then the reservation wage w_bar is set to
    the lowest wage in mcm.w_vec.

    If v(w) < U for all w, then w_bar is set to inf.
    """
    V, U = solve_mccall_model(mcm)
    w_idx = jnp.searchsorted(V - U, 0)
    
    w_bar = jnp.where(w_idx == len(V), jnp.inf, mcm.w_vec[w_idx])
    
    return jax.lax.cond(
        return_values,
        lambda _: (w_bar, V, U),
        lambda _: w_bar,
        None
    )
```

jstac:
I would be in favor of adding type hints for this lecture.  For example, in

```python
def compute_optimal_quantities(c, τ, params, w_vec, p_vec):
```

it would be helpful to know at a glance that params is supposed to be an instance of `EconomyParameters`

HumphreyYang:
Hi @jstac,

Many thanks for the detailed review and for setting the rules! I will apply those rules both to myself and during the review process.

I’ve pushed some changes based on your review:

* removed unnecessary `create_model` functions and redundant functions
* updated `simulate_markov_chain`
* generated the distribution plot using code instead of inserting assets
* added type hints to parameter-bearing instances such as `LakeModel`, `McCallModel`, and `EconomyParameters` so readers can distinguish them

I didn’t change the figure because in

```python
axes[i].plot(x_path[:, i], lw=2, alpha=0.5)
axes[i].hlines(xbar[i], 0, T, 'r', '--')
```

`ax.hlines` restarts the color cycle. So if we don’t explicitly set the color, it ends up using the same blue as the lines from `ax.plot`. That’s why I had to assign a different color to make it visually distinct — even though I realize that goes against the style guide 😬.

> But maybe we should just append to a Python list, or use the `at[i].set(b)` syntax. Thoughts? cc @HumphreyYang @mmcky @cc7768

I definitely think it’s overkill and lacks clarity, so I completely agree with the principle of keeping lectures simple.

To help the discussion, I ran some small experiments:

In the notebook below, I compared the list implementation with the `scan` version on GPU:
[https://colab.research.google.com/gist/HumphreyYang/e42556b0a81f25367695707da4201bcf/for\_scan\_speed.ipynb](https://colab.research.google.com/gist/HumphreyYang/e42556b0a81f25367695707da4201bcf/for_scan_speed.ipynb)

The `for` loop version takes **0.61 seconds**, while the `scan` version takes **0.11 seconds**.

In another notebook, I ran the same comparison on CPU:
[https://colab.research.google.com/gist/HumphreyYang/c385e20d62d021200d49f2cd2b4e33db/for\_scan\_speed.ipynb](https://colab.research.google.com/gist/HumphreyYang/c385e20d62d021200d49f2cd2b4e33db/for_scan_speed.ipynb)

The `for` loop version takes **0.7 seconds**, while the `scan` version takes **0.09 seconds**.

I don’t think speed is really an issue here — but I am slightly concerned about not showing readers the best practice 🤔

Looking forward to hearing your thoughts on the best way to present this!

jstac:
Many thanks @HumphreyYang  for the thoughtful discussion.

Iterating with Claude I thought of the following possibility for handling lectures where it's necessary to generate sequences iteratively.

First, we have a `generate_path` function that handles `lax.scan`.

```python
def generate_path(update_fn, initial_state, num_steps):
    """
    Generate a time series by repeatedly applying an update rule.
    
    This is a general-purpose function that can generate any kind of time series
    given an update rule and initial conditions. It uses JAX's scan function
    to efficiently apply the update rule for the specified number of steps.
    
    Args:
        update_fn: Function that takes (current_state, step_index) and returns 
                  (next_state, current_state). The second return value is stored
                  in the output path.
        initial_state: Starting state for the time series
        num_steps: Number of time steps to simulate
        
    Returns:
        Array containing the full time series path
    """
    # Use JAX scan to efficiently apply update_fn num_steps times
    # scan returns (final_state, full_path) - we only need the path
    _, path = jax.lax.scan(update_fn, initial_state, jnp.arange(num_steps))
    return path
```

The advantage is that we have this logic once and we get to explain it to the reader before we start using it.

Then we have the functions that use this `generate_path` function.  E.g.,

```python
@partial(jax.jit, static_argnames=['T'])
def simulate_stock_path(model, X0, T):
    """
    Simulates the sequence of employment and unemployment stocks.
    
    Args:
        model: Economic model containing parameters for transition matrices
        X0: Initial stock values (employment/unemployment levels)  
        T: Number of time periods to simulate
        
    Returns:
        Array of shape (T, len(X0)) containing stock values at each time step
    """
    # Extract transition matrix from the economic model
    A, A_hat, g = compute_matrices(model)
    
    def stock_update_rule(current_stocks, time_step):
        """Update rule: apply transition matrix to get next period's stocks."""
        next_stocks = A @ current_stocks
        return next_stocks, current_stocks
    
    # Ensure initial state is properly formatted
    initial_stocks = jnp.atleast_1d(X0)
    
    # Generate the stock path using our general path generator
    return generate_path(stock_update_rule, initial_stocks, T)
```

The only thing I don't love about this is that you need to remember the slightly odd structure of the stock_update_rule function -- its nonobvious outputs in particular.  I wonder if some of this logic could be pushed into `generate_path`, so that it could work with a simpler function like

```python
    def stock_update_rule(current_stocks, time_step):
        """Update rule: apply transition matrix to get next period's stocks."""
        next_stocks = A @ current_stocks
        return next_stocks
```

CC @mmcky This discussion is important because it will touch a lot of lectures.  The issue is how to generate time paths in an array processing environment where all arrays are immutable.

Whatever we decide we should probably add it to the style manual.

jstac:
this logic should probably be pushed into `generate_path` as well:

```python
    # Ensure initial state is properly formatted
    initial_stocks = jnp.atleast_1d(X0)
```

HumphreyYang:
> Whatever we decide, we should probably add it to the style manual.

Many thanks, @jstac — this makes a lot of sense to me! I think we can have something like this in the scan update function, wrapping around the single return function to record the carry state:

```python
def scan_wrapper(state, x):
    next_state = update_fn(state, x)
    return next_state, state
```

That way, the `update_fn` can take in the single-return function `stock_update_rule` you suggested.

So the `generate_path` function will look like this:

```python
def generate_path(update_fn, initial_state, num_steps):

    def scan_wrapper(state, x):
        next_state = update_fn(state, x)
        return next_state, state
    
    final_state, path = jax.lax.scan(scan_wrapper, initial_state, jnp.arange(num_steps))
    
    return path
```

This is pretty clean and removes the awkward "two" returns!

I will put this in and update the manual if this looks good to you. I imagine that we only need this in lectures where we have many different sequence generating logic.

jstac:
Yep, that's perfect @HumphreyYang .  Thanks!

I suggest update_wrapper instead of scan_wrapper.   We should be able to jax.jit generate_path with the first argument static.

HumphreyYang:
> I suggest update_wrapper instead of scan_wrapper. We should be able to jax.jit generate_path with the first argument static.

And `num_steps` static : ) I will work on this in this PR!

jstac:
Thanks @HumphreyYang .  Please add in a docstring to both the outer and inner functions -- that explains what it does so we can put it in the docs and then easily copy-paste it to other lectures.

Should `update_fn(state, x)` be `update_fn(state, t)`?  This variable is stepping through the `arange` isn't it?  So it's natural to think of it as a time index.

jstac:
Many thanks @HumphreyYang .  Very nice.  

I slightly reorganized the code so that we can mix it more with discussion.  

Regarding the `rate_steady_state` function, it's not obvious that we're computing an eigenvector.  (In fact I'm not even sure how that works.)  Also, `tol` is not used.  Perhaps it would be best to change to an expliciit eigenvector calculation, in order to match the surrounding discussoin.

jstac:
```
    Returns:
        Array of shape (T, dim(x)) containing the time series path
        [x_0, x_1, x_2, ..., x_{T-1}]
```

Atually the convention is (dim(x), T), so that each column is a state observation x_t.   Let's transpose within the function.

HumphreyYang:
Many thanks @jstac! Your are definitely right. It was a bit too late last night 😬

mmcky:
thanks @HumphreyYang and @jstac -- just catching up with this discussion now the `live` site is fixed. I will use the final solution in this PR to come up with a pattern to add to `QuantEcon.manual`

HumphreyYang:
Hi @jstac and @mmcky, I have pushed changes suggested by @jstac. Please let me know how to improve this template further!

jstac:
This looks great.  Love your work @HumphreyYang

mmcky:
thanks @HumphreyYang and @jstac 

I will merge this PR this afternoon.

jstac:
Thanks @mmcky .   Just a friendly reminder to open an issue for relevant updates to the manual.

mmcky:
thanks @jstac - indeed. Writing that up now before I merge :-) 👍

mmcky:
Screenshot 2025-09-23 at 5 18 48 pm

I have looked into this `Matplotlib` warning. Once we get the new `cache` built -- this should disappear.

PR #609, opened 2025-09-11
[lqcontrol] Proofread and changed text to match the guidelines

matheusvillasb:
- I also added a few lines of text to solve doubts that I had during the reading and that other readers could potentially have as well.
- For the coding part, should I change c_bar for c̄?
- The guideline says not to specify figsize unless necessary. Is this case unnecessary?
- The code seems easy and simple enough to convert to JAX to me. But it would be great to have a second opinion.
- Should I add a comment with the name of the parameters inside the code cells? I've seen this in some lectures, but it's not explicit in the guidelines.

### DIFF
--- a/lectures/bayes_nonconj.md
+++ b/lectures/bayes_nonconj.md
@@ -4,7 +4,7 @@ jupytext:
     extension: .md
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.16.4
+    jupytext_version: 1.17.2
 kernelspec:
   display_name: Python 3 (ipykernel)
   language: python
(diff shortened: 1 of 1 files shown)


mmcky:
@matheusvillasb thanks for this PR. 

The key issue is you have commited the document at the root level rather than replacing `lqcontrol.md` in the `lectures/` folder. 

You should: `git mv lqcontrol.md lectures/lqcontrol.md` and commit. 

You can also just move the file and then git add, git commit as usual.

mmcky:
@matheusvillasb before you make changes to this branch you should

```
git pull origin edits_LQ_lecture
```

when on your `edits_LQ_lecture` branch locally. I just updated it to the latest main.

mmcky:
@matheusvillasb I don't see any difference between `lqcontrol.md` you have committed at the root level and `lectures/lqcontrol.md`

```
(quantecon) ➜  lecture-python.myst git:(edits_LQ_lecture) diff lectures/lqcontrol.md lqcontrol.md
(quantecon) ➜  lecture-python.myst git:(edits_LQ_lecture)
```

HumphreyYang:
Hi @matheusvillasb, many thanks for working on this PR!

> For the coding part, should I change c_bar for c̄?

c̄ isn’t very easy to type in Jupyter notebooks, so we usually only change the Greek letters. For example: `beta_bar` → `β_bar`.

> The guideline says not to specify figsize unless necessary. Is this case unnecessary?

In this case, I think it is necessary to have a stacked plot with the correct aspect ratio. The rule of thumb is whether removing `figsize` makes the graph look worse.

If it doesn’t affect the quality of the figure or the message it conveys, we can remove it.

> The code seems easy and simple enough to convert to JAX to me. But it would be great to have a second opinion.

Yes, I agree. I don’t think we need to convert this. The rule of thumb is to check whether `numba` is being used.

> Should I add a comment with the name of the parameters inside the code cells? I've seen this in some lectures, but it's not explicit in the guidelines.

Yes, if the parameters have specific meanings (for example, economic interpretations). But if they are simply taken from the formula without any extra layer of meaning, it’s fine to leave them as is.

I hope this answers your questions!

mmcky:
@matheusvillasb and I setup a new PR after reverting accidental commits to `main`. The new PR is in #613

PR #621, opened 2025-09-17
[util_rand_resp] Update reference style across lecture and correct typos

HumphreyYang:
This PR aims to update the reference style across the lecture and correct a few errors identified by me and @nisha617.


HumphreyYang:
Hi @mmcky, I think removing the reference in the subtitles fixed the build issue.

(I started to use copilot after your workshop : ))

jstac:

**Netlify preview:** https://pr-621--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 168 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~168 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-621--sunny-cactus-210e3e.netlify.app/util_rand_resp.html
```

---
*This PR touches: `util_rand_resp.md`. Last CI build: `success@2025-10-17`. Branch: 168 commits behind `main` as of 2026-07-08.*

jstac:
Thanks @HumphreyYang — the reference-style conversion and the typo fixes are good, and moving to `mystnb` captions with `numref` is a clear improvement over the hand-numbered "Figure 1.1".

Since you're tied up, I've pushed the changes directly to the branch rather than leaving them as review comments. Summary of what I changed and why:

**One substantive issue.** The plotting cell in the Chaudhuri and Mukerjee section had been replaced by a copy of the Lanke cell — identical except for the caption, the `name:`, and the legend label `Z` → `Q`. That dropped the convex truth border `(√x − 0.4)²` the section is actually about, left `f(x)` as dead code, and moved the optimal point off that border. I've restored the original cell and applied only the intended fix, which was the legend label: it said `Z` while the prose said "point Q".

**Style guide.** A few things pulled back toward the manual:

- Lecture titles take title case (`styleguide/writing.md`); only section headings are sentence case, so the H1 is back to "Expected Utilities of Random Responses". The section-heading changes were right and are kept.
- The two assumptions were nested inside a single `prf:assumption`, so the directive's own "Assumption 1" label collided with the "*Assumption 1*" / "*Assumption 2*" bullets inside it. They're now two directives with titles and labels, following `von_neumann_model.md`, with the second cross-referencing the first.
- `{cite:t}` renders as "Warner [1965]" and can't carry a possessive, so three sentences read as "Warner (1965) original model". Rephrased as "the original model of ``{cite:t}`warner1965randomized` ``". I also converted the two citations the PR missed — Anderson, and the Warner mention in the Overview, which is the worked example in `styleguide/references.md`.
- Captions shortened to the 5–6 word guideline; duplicate `{eq}` reference fixed (`nine-a` appeared twice, second should be `nine-b`).
- Code re-wrapping undone where the original line was already under 79 characters, and backslash continuations removed. The `label=` wraps were justified and are kept.

**Figure pass.** I also took the figures the rest of the way, since they were already being touched:

- Dropped `figsize=(12, 10)` — the guide says don't set figure size without a reason.
- Converted `plt.*` to `fig, ax = plt.subplots()` / `ax.*`, with `lw=2` on line charts.
- In-figure text is roman lower case at the default size ("truth telling", not italic 28pt "Truth Telling"), and legend labels are lower case.

Two calls there worth a look, as they go past mechanical fixes. The iso-variance curves were labelled `V1`…`V9`, which gave a twelve-entry legend that doesn't fit at default size — the family is now labelled once, with the existing "high variance"/"low variance" annotations carrying the ordering, and the prose bullet that referred to `V1`…`V9` reworded to match. And the truth-border legend entries were far too wide, so they're now `$U_i = -p + f(\phi_i)$` etc., with "writing $p$ for $\text{Pr}(A|r_i)$" added to the introducing sentence.

Along the way `Iso_Variance` became a `plot_iso_variance_curves` function (it held two attributes and one plotting method), and `f` is vectorized with `np.where` so the truth border is defined in one place.

All cells execute and I've checked the five figures render without overlap, but the preview is worth a look before merge.

jstac:
Thanks @HumphreyYang and @nisha617  , above comment posted by Claude.  It lists additional style fixes.  To save you time I'll go ahead and merge when this goes green.  Everything looks good on a local build.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/237
**Files synced** (1):
- `lectures/util_rand_resp.md`

PR #624, opened 2025-09-25
[mccall_q] Fix typos, heading case and figure legend

bishmaybarik:
This PR is created to convert the `mccall_q.md` lecture codes into JAX equivalents. 

- [x] The JAX equivalent of all codes should be written
- [x]  Fixing the Headings and Sub-Headings to satisfy Quantecon style guidelines
- [x] Check for grammatical errors and also see if things are easy to follow through
- [x] Adjust the figure legends, as their current position above the curves makes the graphs difficult to interpret.
- [x] Rectify the codes with @HumphreyYang 's suggestions
- [x] Comment out the codes neatly


lectures/mccall_q.md
```diff
@@ -67,22 +67,25 @@ As usual, let's  import some Python modules.
...
-from numba.experimental import jitclass
+import jax
+import jax.numpy as jnp
+import matplotlib.pyplot as plt
+from typing import NamedTuple
 from quantecon.distributions import BetaBinomial
 
-import matplotlib.pyplot as plt
+# set random key for reproducibility
+key = jax.random.PRNGKey(123)
```
HumphreyYang: Let's use `seed=0` as default:

```suggestion
key = jax.random.PRNGKey(0)
```

lectures/mccall_q.md
```diff
@@ -146,81 +149,78 @@ Next we'll compute the worker's optimal value function by iterating to convergen
...
+      def cond_fun(state):
+          v, i, error = state
+          return (error > eps) & (i < max_iter)
+
+      # Initial state: (v, iteration, error)
+      init_state = (v_init, 0, eps + 1)
+      final_v, final_i, final_error = jax.lax.while_loop(cond_fun, body_fun, init_state)
+
+      flag = jnp.where(final_error <= eps, 1, 0)
+      return final_v, flag
```
HumphreyYang: In this code we are resetting `v_next` using `for` loop many times. I think it is not the best practice in `jax`.

One way to rewrite is using vectorized map like this:

```python
@jax.jit
def update(model, v):
    n = model.w.shape[0]

    def v_at_state(i):
        sa = state_action_values(model, i, v)
        return jnp.max(sa)

    indices = jnp.arange(n)
    v_new = jax.vmap(v_at_state)(indices)
    return v_new

@jax.jit
def vfi(model, tol=1e-5, max_iter=500):

    v0 = model.w / (1.0 - model.β)

    def body_fun(state):
        v, i, err = state
        v_new = update(model, v)
        err_new = jnp.max(jnp.abs(v_new - v))
        return v_new, i + 1, err_new

    def cond_fun(state):
        _, i, err = state
        return (err > tol) & (i < max_iter)

    init_state = (v0, 0, tol + 1.0)
    v_final, iters, err = jax.lax.while_loop(cond_fun, body_fun, init_state)
    converged = jnp.where(err <= tol, 1, 0)
    return v_final, converged
```

One note on the code above is that I changed the function name to lower case and changed `flag` to `converged` which is closer to the meaning of the variable. 

This code is also more scalable.

I will leave the code commenting to you : )

lectures/mccall_q.md
```diff
@@ -146,81 +149,78 @@ Next we'll compute the worker's optimal value function by iterating to convergen
...
+          * ax is an axes object that implements a plot method.
+
+      """
+      n = len(mcm.w)
+      v = mcm.w / (1 - mcm.β)
+      v_next = jnp.empty_like(v)
+      for i in range(num_plots):
+          ax.plot(mcm.w, v, '-', alpha=0.4, label=f"iterate {i}")
+          # Update guess
+          for j in range(n):  # changed variable name to avoid conflict
```
HumphreyYang:
```suggestion
          for j in range(n):
```

lectures/mccall_q.md
```diff
@@ -146,81 +149,78 @@ Next we'll compute the worker's optimal value function by iterating to convergen
...
+          v = v_next # handling immutability
+
+      ax.legend(loc='lower right')
 ```
 
 ```{code-cell} ipython3
-mcm = McCallModel()
-valfunc_VFI, flag = mcm.VFI()
+mcm = create_mccall_model()
+valfunc_VFI, flag = VFI(mcm)
```
HumphreyYang: please avoid using capital letters in the variable name and function name. See PEP8 rule here: https://peps.python.org/pep-0008/#function-and-variable-names

Same holds  for the variable names below.

lectures/mccall_q.md
```diff
@@ -488,127 +488,158 @@ For simplicity and convenience, we let `s` represent the state index between $0$
...
-            if self.quit_allowed == 0:
-                TD = w[state_next] + β*np.max(qtable[state_next, :]) - qtable[state, accept]
-            else:
-                TD = w[state_next] + β*qtable[state_next, 1] - qtable[state, accept]
+class QlearningMcCall(NamedTuple):
+    c: float                  # unemployment compensation
+    β: float                  # discount factor
+    w: jnp.ndarray            # array of wage values, w[i] = wage at state i
+    q: jnp.ndarray            # array of probabilities
+    eps: float                # for epsilon greedy algorithm
```
HumphreyYang: Same for the variable `eps` and `epsilon` below

```suggestion
    ε: float                # for ε-greedy algorithm
```

lectures/mccall_q.md
```diff
@@ -488,127 +488,158 @@ For simplicity and convenience, we let `s` represent the state index between $0$
...
+                            quit_allowed=0):
+    return QlearningMcCall(c=c,
+                           β=β,
+                           w=w,
+                           q=q,
+                           eps=eps,
+                           δ=δ,
+                           lr=lr,
+                           T=T,
+                           quit_allowed=quit_allowed)
```
HumphreyYang: In this case, we can just put default values into the class because we don’t need to perform operations on them.

We only need functions like `create_qlearning_mccall` when we need to derive some of the values from others.

lectures/mccall_q.md
```diff
@@ -488,127 +488,158 @@ For simplicity and convenience, we let `s` represent the state index between $0$
...
 
-    def run_one_epoch(self, qtable, max_times=20000):
-        """
-        Run an "epoch".
-        """
+    def accept_case():
+        state_next = state
+        TD = jnp.where(model.quit_allowed == 0,
+                       w[state_next] + β * jnp.max(qtable[state_next, :]) - qtable[state, accept],
+                       w[state_next] + β * qtable[state_next, 1] - qtable[state, accept])
```
HumphreyYang: Please try to keep each line under 80 characters.

lectures/mccall_q.md
```diff
@@ -488,127 +488,158 @@ For simplicity and convenience, we let `s` represent the state index between $0$
...
+
+
+def run_epochs(N, qlmc, qtable, key):
     """
     Run epochs N times with qtable from the last iteration each time.
     """
-
     for n in range(N):
-        if n%(N/10)==0:
+        if n % (N // 10) == 0:
```
HumphreyYang: I think `copilot` is right about these lines. Also, they are not following PEP 8 : )

lectures/mccall_q.md
```diff
@@ -146,81 +149,78 @@ Next we'll compute the worker's optimal value function by iterating to convergen
...
+      """
+      n = len(mcm.w)
+      v = mcm.w / (1 - mcm.β)
+      v_next = jnp.empty_like(v)
+      for i in range(num_plots):
+          ax.plot(mcm.w, v, '-', alpha=0.4, label=f"iterate {i}")
+          # Update guess
+          for j in range(n):  # changed variable name to avoid conflict
+              v_next = v_next.at[j].set(jnp.max(state_action_values(mcm, j, v)))
+          v = v_next # handling immutability
```
HumphreyYang: Here we can use 

```python
indices = jnp.arange(n)
v_next = jax.vmap(
    lambda j: jnp.max(state_action_values(mcm, j, v)))(indices)
v = v_next
```

lectures/mccall_q.md
```diff
@@ -488,127 +488,158 @@ For simplicity and convenience, we let `s` represent the state index between $0$
...
+class QlearningMcCall(NamedTuple):
+    c: float                  # unemployment compensation
+    β: float                  # discount factor
+    w: jnp.ndarray            # array of wage values, w[i] = wage at state i
+    q: jnp.ndarray            # array of probabilities
+    eps: float                # for epsilon greedy algorithm
+    δ: float                  # Q-table threshold
+    lr: float                  # the learning rate α
+    T: int                    # maximum periods of accepting
+    quit_allowed: int          # whether quit is allowed after accepting the wage offer
```
HumphreyYang: This is over 80 charaters. We can write 

```suggestion
    quit_allowed: int          # whether quit is allowed
```

lectures/mccall_q.md
```diff
@@ -651,47 +682,48 @@ ax.set_xlabel('wages')
...
-        qtable = new_qtable
-
-    ax.set_xlabel('wages')
-    ax.set_ylabel('optimal value')
-    ax.legend(loc='lower right')
-    plt.show()
+def plot_epochs(epochs_to_plot, quit_allowed=1, key=None):
+      "Plot value function implied by outcomes of an increasing number of epochs."
+      if key is None:
+          key = jax.random.PRNGKey(42)  # Default key if none provided
```
HumphreyYang: We defined `key` at the beginning of the lecture, so we can remove these lines and use the `key` as input.

HumphreyYang (commented):
Many thanks @bishmaybarik! Nice changes -- they are really good improvements. Here are some minor comments.

bishmaybarik:
hi @HumphreyYang , these are extremely useful suggestions -- thanks a lot! I'll make all the necessary changes and push the updated version once ready.

bishmaybarik:
hi @HumphreyYang and @mmcky , may I know if you have any thoughts on these updates? It would be great to hear from you!

jstac:

**Netlify preview:** https://pr-624--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 175 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~175 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-624--sunny-cactus-210e3e.netlify.app/mccall_q.html
```

---
*This PR touches: `mccall_q.md`. Last CI build: `success@2025-10-06`. Branch: 175 commits behind `main` as of 2026-07-08.*

jstac:
From @bishmaybarik :

hi @HumphreyYang and @mmcky , may I know if you have any thoughts on these updates? It would be great to hear from you!

From me: I can handle this if you don't have time, just let me know.

jstac:
Thanks @bishmaybarik, and apologies for the slow review — this sat much longer than it should have.

I've taken the branch in a different direction rather than let it go stale, so I want to explain the reasoning.

**On the JAX conversion.** I don't think `mccall_q` should be a JAX lecture. Q-learning here is inherently sequential: `run_one_epoch` walks up to 20,000 steps, each reading and writing a single cell of an 11×2 Q-table based on the step before it. There's nothing to vectorize and nothing to parallelize. That's the case the style guide addresses directly under [Sequential: Numba wins](https://manual.quantecon.org/styleguide/jax.html):

> For **inherently sequential** work that does not need autodiff, prefer **Numba**. At best JAX ties on speed — and only after being pinned to the CPU — while the syntax is considerably less clear.

The branch itself makes the point: it sets `jax.config.update('jax_platform_name', 'cpu')` in the imports cell. This repo's CI runs on a GPU instance with `jax[cuda12-local]`, so the conversion moves the lecture to JAX and then switches off the hardware that would justify it. That's a reliable signal the algorithm isn't a fit.

This is a judgment call about the lecture, not about the work — the conversion itself is competent, and the `NamedTuple` + factory-function structure is exactly what the guide asks for. It's just applied to a lecture that shouldn't have been on the conversion list.

**One thing worth knowing for future conversions.** There's a bug in the converted `run_one_epoch` that's easy to miss:

```python
def cond_fun(state):
    qtable, s, accept_count, t, key = state
    error = jnp.where(t==0, δ + 1, jnp.max(jnp.abs(qtable - state[0])))
```

`state[0]` is `qtable` — unpacked from it on the line above — so `error` is identically zero for every `t > 0` and the loop exits after one iteration. I ran it to confirm: each "epoch" performs exactly one TD update instead of running to convergence. The Numba original compared the pre- and post-update tables (`qtable_new - qtable`); that comparison got lost in the rewrite. Worth watching for when converting a `while` loop whose exit test depends on the *change* in the carry — the carry has to hold both the old and new values, or the test has to move into the body.

**What I kept.** The branch had real improvements independent of JAX, and those are now applied on top of main:

- all the typo and wording fixes — "illegitmate", "previos", "prematurally", "algorthm", "objection to reject" → "option to reject", "$Q$ map" → "$Q$ maps"
- section headings to sentence case, per the style guide
- the legend fix in `plot_epochs` — it now sits below the axes in two columns instead of covering the curves. I also formatted the mean error to two decimals, since the full-precision labels were too wide to fit
- the duplicated VFI computation at the end of the wage-distribution cell, which ran again in the very next cell

Main had also moved these functions to the `np.random.default_rng` Generator API in #959, so I merged main in to resolve that. Your commits are still in the branch history.

Retitled accordingly. Thanks again for the work here — the reference material is in the style guide's [when to use JAX](https://manual.quantecon.org/styleguide/jax.html) checklist if you're picking up another conversion.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/236
**Files synced** (1):
- `lectures/mccall_q.md`

PR #644, opened 2025-10-10
[kalman_2] Use a typed NamedTuple for WorkerModel

xuanguang-li:
This pull request improves the `lectures/kalman_2.md` file, focusing on code modernization, clarity, and consistency.  
The main updates include:

* Refactored the code to use `jax` for array creation, mathematical operations, and simulation functions. Some `numpy` code remains to avoid data type–related warnings.  
* Improved the logic of the Kalman filter implementation, including clarifying the meaning of the variable `x_hat` and handling the initial states more carefully.  
* Standardized parameter names in worker creation and simulation functions — for example, replacing `hhat_0` and `uhat_0` with `μ_h` and `μ_u`, and defining `hhat_0` and `uhat_0` as the priors.  
* Added a `seed` parameter to all simulation calls.  
* Fixed typos and added explanatory comments for better readability.


mmcky:
thanks @xuanguang-li. I have just merged in the latest `main` and the preview is now showing `kalman_2` as the change.

xuanguang-li:
Thanks @mmcky. I noticed a warning right below the import cell, saying.

> Matplotlib is building the font cache; this may take a moment.

Should we hide this output or leave it there?

jstac:

**Netlify preview:** https://pr-644--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 174 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~174 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-644--sunny-cactus-210e3e.netlify.app/kalman_2.html
```

---
*This PR touches: `kalman_2.md`. Last CI build: `success@2025-10-15`. Branch: 174 commits behind `main` as of 2026-07-08.*

xuanguang-li:

jstac:
Thanks @xuanguang-li, and sorry this sat for so long — that delay is the source of most of what follows.

I've reworked the branch rather than let it go stale further, so here's what changed and why.

**Dropping the JAX conversion.** The style guide names this case almost directly, under [when not to use JAX](https://manual.quantecon.org/styleguide/jax.html):

> Substituting `jnp` for `np` is *not* converting a lecture to JAX. If the numerical work still runs through SciPy, `scipy.stats`, or NumPy-based QuantEcon.py classes such as **`Kalman` and `LinearStateSpace`**, then the arrays are merely being shuttled between host and device at every call. The lecture gets JAX's constraints and none of its benefits.

All the actual computation in `kalman_2` stays inside `quantecon.Kalman`, `LinearStateSpace`, and `scipy.stats.multivariate_normal` — only the array constructors moved to `jnp`, which the PR description acknowledges ("Some `numpy` code remains to avoid data type–related warnings"). Two related items followed from that: the GPU admonition was added to a lecture with no GPU work (and pasted inline rather than via `{include} _admonition/gpu.md`), and `x_hat.at[:, t].set(...)` appears inside a Python `for` loop in three places, which is $O(T^2)$ outside `jit`. Both are listed as anti-patterns in the guide.

None of this is a comment on the quality of the work — it's a judgement about the lecture. The conversion is competently done; it's applied to a lecture where JAX has nothing to bite on.

**Rebasing onto main's current text.** This is the bigger issue, and it isn't your doing. The branch was opened in October 2025, and `kalman_2.md` has been substantially rewritten on main four times since:

- #903 (2026-06-15) — exposition of the first three sections
- #907 (2026-06-16) — code and descriptions
- #931 (2026-06-22) — Tom's Kalman filter lectures
- #944 (2026-07-06) — style guide alignment

The automated conflict resolution in `1f3225b` resolved in favour of the branch, so the diff had quietly become a revert of all four. Among the things that would have been lost: the derivation of the Kalman gain as $K_t = A\Sigma_t G'(G\Sigma_t G' + R)^{-1} = A L_t$ with the explanation of why conditioning on $y^{t-1}$ makes $K_t$ combine the filtering update with the one-step-ahead prediction; the discussion of why learning speed varies with $g\beta/(1-\alpha)$; the observability conditions ($\beta = 0$ or $g = 0$ and the firm can never learn $u_0$); and the fix making `Σ_0` hold $\sigma^2$ rather than $\sigma$.

There was also a subtle regression worth flagging, since it's the kind of thing easy to lose in a large rewrite: in the fixed-initial-state section the branch reassigned `worker = create_worker(μ_h=0.0, μ_u=4.0)` and then passed `worker.μ_0` in as the firm's prior. Main deliberately keeps that in a separate `worker_example` so the firm's prior stays at $(4, 4)$ — otherwise the firm's prior *is* the truth and there's nothing left for the filter to learn, which is the point of the section. A couple of smaller ones: `(kalman_2)=` had been changed to `(kalman)=`, which collides with the existing target in `kalman.md`; and `ss_μ = jnp.array([[1], [2]])` sets $h_0=1, u_0=2$ under a comment saying the opposite.

So I merged current `main` and took its version of the lecture, then reapplied your one clearly additive change on top.

**What's kept.** `WorkerModel` is now a typed `typing.NamedTuple` rather than a `collections.namedtuple`, alongside the existing `create_worker` factory — the model-structure pattern the style guide asks for. The prose references were updated to match, including "our `create_worker` namedtuple" → "factory function", which was inaccurate before. The PR is now +12/−6 and retitled accordingly. Your commits remain in the branch history.

**Where the JAX idea does have a future.** Worth saying, because I don't think the instinct was wrong — just early. A linear-Gaussian state-space model is a genuinely differentiable object: the log-likelihood is smooth in $(\alpha, \beta, c, R, g)$, and the style guide treats differentiability as a legitimate reason to lean toward JAX. What blocks it here is that `jax.grad` cannot see through `quantecon.Kalman`. Write the filter recursion itself in JAX — a `lax.scan` over the predict/update steps — and the lecture gains something it currently cannot do at all: differentiate the likelihood and estimate the worker's parameters by gradient ascent. That's a real lecture rather than a conversion, and `kalman_filter_var` may be its more natural home since it already builds innovations and likelihoods from the same recursion. If you're interested in picking that up, it'd be a good one.

jstac:
Thanks @xuanguang-li , comments above from Claude.  Sorry to be so slow getting to this. I'll merge when green.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/238
**Files synced** (1):
- `lectures/kalman_2.md`

PR #645, opened 2025-10-13
[harrison_kreps]: Fix axis

UdohNakamura:
- Problem: np.max(..., axis=1) / np.min(..., axis=1) compares across states, not across investor types.
- Spec: Per equations (78.3) and (78.5), the max/min must be taken across types → axis=0.
- Fix: Replace axis=1 with axis=0 in the three lines below.

### DIFF
--- a/lectures/harrison_kreps.md
+++ b/lectures/harrison_kreps.md
@@ -414,7 +414,7 @@ def price_optimistic_beliefs(transitions, dividend_payoff, β=.75,
         p_old = p_new
         p_new = β * np.max([q @ p_old
                             + q @ dividend_payoff for q in transitions],
-                            1)
+                            axis=0)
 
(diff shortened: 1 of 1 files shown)


HumphreyYang:
Many thanks @UdohNakamura! This looks great to me.

Hi @mmcky, this is a small (but important) fix. Should we merge it when you are ready? @UdohNakamura's change does not affect output because the two agent case is pretty special (with the same off diagnoal entries being the smallest ones). If we add a three-state two-agent exercise then it will cause issue.

mmcky:
thanks @UdohNakamura -- there seems to be an issue with permissions to `netlify` b/c this originates from a fork. 

@HumphreyYang the logs indicate this is building well so I'll merge now.

PR #652, opened 2025-10-21
[two_auctions] Fix conditional expectation simulation

oyamad:
## Before
image

Looking at such a figure which exhibits a systematic discrepancy, one should suspect that there should be a bug somewhere.

The values in the array `v` in this line https://github.com/QuantEcon/lecture-python.myst/blob/33416c7da68e67aa04213f6b218b953533c66db0/lectures/two_auctions.md?plain=1#L434 have already been sorted. One has to redraw random valuations to pass to `evaluate_largest`.

## After
image

### DIFF
--- a/lectures/two_auctions.md
+++ b/lectures/two_auctions.md
@@ -424,11 +424,11 @@ def evaluate_largest(v_hat, array, order=1):
 
 We can check the accuracy of our `evaluate_largest` method by comparing it with an analytical solution.
 
-We find that the evaluate_largest method functions well.
+We find that the `evaluate_largest` method functions well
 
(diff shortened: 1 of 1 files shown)


mmcky:
thank you @oyamad

mmcky:
@HumphreyYang this looks good. I will merge this later today if you want to have a quick look.

HumphreyYang:
Many thanks @oyamad and @mmcky. It looks great to me! I just pushed a minor edit to add some spaces that is unrelated to @oyamad's edit.

Please feel free to merge it and make it live (as well as the Morris lecture).

PR #654, opened 2025-10-22
[markov_asset] Update to JAX and runtime comparison

xuanguang-li:
This pull request modernizes the `markov_asset.md` notebook by refactoring the codebase to use JAX and JAX NumPy.

- Refactored all numerical computations and linear algebra from `numpy` and `numpy.linalg` to `jax.numpy` and `jax.numpy.linalg`. 
- Replaced the mutable `AssetPriceModel` class with an immutable, type-safe `NamedTuple` version and introduced a `create_ap_model` factory function for consistent construction.
- Updated function signatures and internal logic to use the new model structure. 
- Replaced manual Python loops with JAX functional control flow constructs such as `jax.lax.while_loop` and `jax.lax.fori_loop`.
- Fixed minor bugs in parameter initialization, state handling, and tolerance defaults. 

### Update: Runtime Comparison Between `JAX (GPU)`, `JAX (CPU)`, and `NumPy`

**Methodology:**
- The `JAX` version uses the code in this PR, while the `NumPy` version uses the code in `main`.
- The runtime for `JAX (GPU)` is measured using Google Colab.
- Runtime is collected using `qe.timeit` over 1,000 iterations.
- Each iteration consists of one function that solves the consol price and one that solves the call option price.

**Results:**
- Average runtime: `JAX (CPU)` > `NumPy` > `JAX (GPU)`.
- More details are shown in the attached plots.

runtime_compare_average
runtime_compare_boxplot


jstac:

**Netlify preview:** https://pr-654--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 164 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~164 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-654--sunny-cactus-210e3e.netlify.app/markov_asset.html
```

---
*This PR touches: `markov_asset.md, markov_asset_numpy.md`. Last CI build: `failure@2026-07-07`. Branch: 164 commits behind `main` as of 2026-07-08.*

lectures/markov_asset.md
```diff
@@ -838,36 +903,58 @@ def call_option(ap, ζ, p_s, ϵ=1e-7):
...
+    M = P * G**(- γ)
 
     # Make sure that a unique consol price exists
-    ap.test_stability(M)
+    test_stability(M, β)
 
     # Compute option price
     p = consol_price(ap, ζ)
-    w = np.zeros(ap.n)
+    err.throw()
```
jstac, line 914: This seems to be added by mistake.

jstac:
Many thanks @xuanguang-li , much appreciated.

By the way, you added a file to this PR that I assume is a mistake: https://github.com/QuantEcon/lecture-python.myst/pull/654/changes#diff-200962c576f4fdd31f65eea1717388dafb176a3bb5ec8630dbe0840a61f49f57

You can leave this PR with me.  I'll add any necessary fixes and then merge it.

jstac:
Hi @xuanguang-li .  (CC @mmcky @HumphreyYang )

I've now reviewed this carefully, including testing with Claude on my CPU and GPU.  

I always ask Claude to read our current style guide before assessing.  Here it recommends keeping the lecture in NumPy and I agree.  JAX is slower on the GPU, and only somewhat faster if we pin it to using the CPU.

This comes at some readability cost, with the pin to the CPU, lax.while loop, etc.

For that reason, I've decided to close the PR without merging.  This lecture can stay in NumPy.

FYI, here is Claude's report:

  The GPU is ~5–6× slower than NumPy for this lecture, plus a ~0.3 s one-time compile hit. Root cause: call_option's jax.lax.while_loop
  runs 182 iterations to converge, and each iteration is a tiny 25×25 matmul + a scalar convergence check. That's a latency-bound, 
  sequential workload — 182 back-to-back GPU kernel launches with device round-trips per iteration. GPUs are throughput devices; this is
  close to their worst case. NumPy runs all 182 iterations in-process with no launch overhead.

  This resolves the apparent contradiction with the PR description (which reported GPU as fastest): that benchmark was almost certainly
  measuring JAX's async dispatch, not completed compute — a classic JAX benchmarking trap where you time how fast work is queued rather
  than finished. Once you sync the device, the GPU advantage disappears and reverses.

  Bottom line: this empirically confirms style-guide point #4 — markov_asset is not computationally intensive (tiny 25×25 / 5×5
  matrices), so JAX/GPU delivers no benefit and actively hurts (compile latency + slower steady-state). The author's "very long runtime
  on GPU" is real, expected, and inherent to the problem size, not a bug to fix.

  My recommendation to you

  I'd lean toward not converting this lecture to JAX at all — keep it NumPy per the style guide's explicit carve-out — rather than
  merging a slower, less-readable (checkify-heavy) version. If the migration tracker (meta#299) mandates conversion regardless, then at
  minimum items 1–3 from the saved response still need fixing first.

xuanguang-li:
Thanks, @jstac, for your diagnosis. 

The large overhead charged in GPU is impressive. This will help us to understand the trade-off between running JAX code on GPU and CPU.

PR #656, opened 2025-10-26
[hoist_failure] Clean up the code and correct minor errors

HumphreyYang:
This PR cleans up the code and correct minor errors and typos across the lecture.

It also improves the printing and smooth some of the sentences.


jstac:

**Netlify preview:** https://pr-656--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 166 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~166 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-656--sunny-cactus-210e3e.netlify.app/hoist_failure.html
```

---
*This PR touches: `hoist_failure.md`. Last CI build: `success@2025-11-14`. Branch: 166 commits behind `main` as of 2026-07-08.*

Chihiro2000GitHub:
Hi @HumphreyYang, I'm working through the NumPy random API migration in QuantEcon/meta#299, and `hoist_failure.md` is one of the lectures still on the list.

This PR already rewrites the sampling cell, so rather than open a separate PR that changes the same lines, I thought I would ask here first. While you are in there, would you be happy to use the Generator API for those draws?

```python
rng = np.random.default_rng(1234)
s1 = rng.lognormal(μ, σ, n_samples)
s2 = rng.lognormal(μ, σ, n_samples)
s3 = rng.lognormal(μ, σ, n_samples)
```

That keeps the seed you added, and it saves us both from a conflict on the same lines.

If you would rather keep this PR as it is, I am happy to do the migration in a separate PR once this one is merged.

jstac:
Thanks @Chihiro2000GitHub, and sorry for the slow reply — @HumphreyYang hasn't been able to get back to this one, so I've picked it up.

Yes, let's do it here rather than in a separate PR. I've made exactly the change you suggested on this branch:

```python
rng = np.random.default_rng(1234)
s1 = rng.lognormal(μ, σ, n_samples)
s2 = rng.lognormal(μ, σ, n_samples)
s3 = rng.lognormal(μ, σ, n_samples)
```

It replaces the `np.random.seed(1234)` plus legacy `np.random.lognormal` that this branch had introduced, and keeps the seed. That's the only sampling in the lecture, so `hoist_failure.md` should be clear for meta#299 once this merges — worth ticking off your list rather than opening a follow-up.

Asking first was the right call: this branch rewrites that cell substantially, so a parallel PR would have conflicted on exactly those lines.

While updating the branch I also merged current `main` into it (it was 218 commits behind, which was why the preview was 404ing), added the `mystnb` captions the figures style guide asks for, and brought the install cell into line with the other cells. The figures will shift slightly because PCG64 draws a different stream from the legacy `RandomState` — shape and magnitude are unchanged, which is the expected outcome noted in the migration guidance.

jstac:
Opened #1027 for the two new exercises — exercise 2 can't be answered as posed (the comparison it asks for holds by linearity of expectation regardless of the rare event approximation), and exercise 1's conclusion of $p = 13$ disagrees with the $p = 15$ used in the lecture body. Neither is a regression against `main`, so this PR needn't be blocked on them.

Chihiro2000GitHub:
Thank you @jstac for picking this up!

I really appreciate you making the change I suggested -- that's very helpful. I've noted this in https://github.com/QuantEcon/meta/issues/299.

Thanks again!

PR #687, opened 2025-11-10
Fix JAX compatibility issues in Job Search III lecture

jstac:
## Summary

This PR fixes several JAX-related compatibility issues in the **Job Search III: Search with Separation and Markov Wages** lecture (`mccall_model_with_sep_markov.md`).

### Main Fixes
- **Refactored `vfi()` function**: Now returns only `v_final` instead of a tuple `(v_star, σ_star)`, making the implementation cleaner and more consistent with typical VFI patterns
- **Removed `successive_approx()` function**: Integrated the iteration logic directly into `vfi()` to simplify the codebase
- **Fixed `get_reservation_wage()` function**: Rewrote to use `jnp.argmax()` instead of `jnp.where()` to avoid JAX concretization errors that occur during JIT compilation
- **Fixed JAX decorator imports**: 
  - Changed `@jit` to `@jax.jit` 
  - Changed `@partial(jit, ...)` to `@partial(jax.jit, ...)`
  - Removed unnecessary `@jit` decorators from `T()` and `get_greedy()` functions
- **Updated all `vfi()` call sites**: Added explicit `σ_star = get_greedy(v_star, model)` calls after each `vfi()` invocation (7 locations total)

### Additional Improvement
- Minor wording improvement in `mccall_model_with_separation.md` for better clarity

## Testing
✅ Converted the markdown file to Python using `jupytext`  
✅ Ran the Python script successfully without errors  
✅ Verified all MyST markdown structure is valid  
✅ Confirmed jupytext can parse and convert the file

## Technical Details

The main issue was that `jnp.where()` with dynamic size cannot be used inside JIT-compiled functions, causing a `ConcretizationTypeError`. The fix uses `jnp.argmax()` to find the first acceptance index, which works correctly with JAX's tracing mechanism.

### DIFF
--- a/lectures/mccall_model_with_sep_markov.md
+++ b/lectures/mccall_model_with_sep_markov.md
@@ -509,7 +509,7 @@ This holds because:
 
 These properties ensure the chain is ergodic with a unique stationary distribution $\pi$ over states $(s, w)$.
 
-For an ergodic Markov chain, the ergodic theorem guarantees that time averages = ensemble averages.
+For an ergodic Markov chain, the ergodic theorem guarantees that time averages = cross-sectional averages.
 
 In particular, the fraction of time a single agent spends unemployed (across all
 wage states) converges to the cross-sectional unemployment rate:
@@ -661,7 +661,12 @@ def plot_cross_sectional_unemployment(model: Model, t_snapshot: int = 200,
     plt.show()
 ```
 
-Now let's compare the time-average unemployment rate (from a single agent's long simulation) with the cross-sectional unemployment rate (from many agents at a single point in time):
+Now let's compare the time-average unemployment rate (from a single agent's long simulation) with the cross-sectional unemployment rate (from many agents at a single point in time).
+
+We claimed above that these numbers will be approximately equal in large
+samples, due to ergodicity.
+
+Let's see if that's true.
 
 ```{code-cell} ipython3
 model = create_js_with_sep_model()
@@ -677,20 +682,23 @@ print(f"Cross-sectional unemployment rate (at t=200): "
 print(f"Difference: {abs(time_avg_unemp - cross_sectional_unemp):.4f}")
 ```
 
+Indeed, they are very close.
+
 Now let's visualize the cross-sectional distribution:
 
 ```{code-cell} ipython3
 plot_cross_sectional_unemployment(model)
 ```
 
-## Cross-Sectional Analysis with Lower Unemployment Compensation (c=0.5)
+## Lower Unemployment Compensation (c=0.5)
 
-Let's examine how the cross-sectional unemployment rate changes with lower unemployment compensation:
+What happens to the cross-sectional unemployment rate with lower unemployment compensation?
 
 ```{code-cell} ipython3
 model_low_c = create_js_with_sep_model(c=0.5)
 plot_cross_sectional_unemployment(model_low_c)
 ```
+
 
 ## Exercises
 
@@ -698,7 +706,7 @@ plot_cross_sectional_unemployment(model_low_c)
 :label: mmwsm_ex1
 ```
 
-Create a plot that shows how the steady state cross-sectional unemployment rate
+Create a plot that investigates more carefully how the steady state cross-sectional unemployment rate
 changes with unemployment compensation.
 
 ```{exercise-end}


jstac:
## Key Fix: get_reservation_wage() Function

The most critical fix in this PR addresses a JAX concretization error. The original implementation used:

```python
accept_indices = jnp.where(σ == 1)[0]
if len(accept_indices) == 0:
    return jnp.inf
return w_vals[accept_indices[0]]
```

This fails in JIT-compiled functions because:
1. `jnp.where()` requires a statically-known size when used with JAX tracing
2. The `if len(...)` check tries to concretize an abstract value

The new implementation uses:
```python
first_accept_idx = jnp.argmax(σ)
return jnp.where(jnp.any(σ), w_vals[first_accept_idx], jnp.inf)
```

This works because:
- `jnp.argmax()` finds the first True value without requiring dynamic sizing
- `jnp.where()` is used with scalar condition (no dynamic size issues)
- Everything remains traceable through JAX's JIT compilation

jstac:
## Refactoring Notes: vfi() Function

The `vfi()` function was simplified:

**Before:**
- Used a separate `successive_approx()` helper function
- Returned both value function and policy: `return v_star, σ_star`

**After:**
- Iteration logic integrated directly into `vfi()`
- Returns only value function: `return v_final`
- Policy computed separately via `σ_star = get_greedy(v_star, model)`

**Benefits:**
1. **Cleaner separation of concerns**: VFI computes values, policy extraction is separate
2. **More flexible**: Users can compute value function without necessarily needing the policy
3. **Consistent with typical VFI patterns**: Most VFI implementations return just the value function
4. **Less code**: Removed 35 lines of the `successive_approx()` function

All 7 call sites in the lecture were updated to explicitly call `get_greedy()` after `vfi()`.

jstac:
## Testing Details

The changes were thoroughly tested:

1. **Conversion test**: Used `jupytext --to py` to convert the markdown to Python
2. **Execution test**: Ran the Python script successfully with output:
   ```
   Time-average unemployment rate (single agent): 0.2215
   Cross-sectional unemployment rate (at t=200): 0.2257
   Difference: 0.0042
   ```
3. **Round-trip test**: Converted back to markdown using `jupytext --to myst`
4. **Validation checks**:
   - ✅ 21 code cells all properly formatted
   - ✅ All MyST directives balanced (exercise/solution blocks)
   - ✅ Math delimiters balanced (10 `$$` blocks)
   - ✅ YAML frontmatter valid
   - ✅ Jupytext can parse without errors

All fixes are backwards-compatible in terms of functionality - the lecture produces the same numerical results and plots as before.

jstac:
## Standardized continuation value notation

I've updated the Job Search II lecture (McCall model with separation) to use more consistent notation with the basic McCall model.

### Changes made:
- **Changed continuation value from `d` to `h`**: Now using `h = u(c) + β * sum_w v_u(w) q(w)` instead of `d = sum_w v_u(w) q(w)`
- This mirrors the basic model's `h = c + β * sum_w v*(w) q(w)`, making the progression between lectures more intuitive
- The notation is cleaner since `h` now directly represents the value of rejecting an offer

### Benefits:
- **Better consistency**: Readers can more easily connect concepts between Job Search I and II
- **Clearer mathematics**: The continuation value `h` explicitly shows both current utility `u(c)` and discounted future value
- **Simpler equations**: Bellman equation becomes `v_u(w) = max{v_e(w), h}` instead of `max{v_e(w), u(c) + β*d}`

All code has been updated and tested successfully. The mathematical derivations and implementation now align better with the pedagogical flow of the lecture series.

mmcky:
@jstac this quick fix looks good to go. All execution is passing and the two new lectures are building plots etc. with no Warnings.

jstac:
Thanks @mmcky !!  I'll check the lectures and merge if they are all good.

jstac:
Are we still using pytorch btw @mmcky ?

ripgrep says no ...

jstac:
Ah, sorry, I see you just removed it!

@mmcky

mmcky:
> Are we still using pytorch btw @mmcky ?
> 
> ripgrep says no ...

@jstac we use `numpyro` which uses pytorch (under the hood I believe) -- but it should install binaries.
 I will run a full test build after merge for the cached lectures.

jstac:
Thanks @mmcky 

Another reason to break up these big lecture series, isn't it?

The issues in the build were nothing to do with my PRs.  I had a pretty stressful morning trying to get the new lectures live and eventually giving up :grimacing:

mmcky:
Sorry for your morning @jstac. Infrastructure / install issues with JAX are stressful! 

I am about to run a full test build on another branch -- but there must have been a nightly release of pytorch that has a pretty significant change in it.

jstac:
> > Are we still using pytorch btw @mmcky ?
> > ripgrep says no ...
> 
> @jstac we use `numpyro` which uses pytorch (under the hood I believe) -- but it should install binaries. I will run a full test build after merge for the cached lectures.

@mmcky this is from perplexity:

Yes, you can use NumPyro without PyTorch. NumPyro is a lightweight probabilistic programming library that uses JAX as its backend for automatic differentiation and JIT compilation. It does not depend on PyTorch at all. In fact, NumPyro is designed to leverage JAX's NumPy-compatible API and supports running on CPU, GPU, or TPU through JAX.

mmcky:
Sorry @jstac used the wrong name, should be `pyro-ppl` -- we install `numpyro` and `pyro-ppl` which was previously used. Just running a search to see when we removed it. I actually think @HumphreyYang removed the use of `pyro-ppl` fairly recently.

jstac:
## Additional updates to `mccall_model_with_separation.md`

I've made several improvements to the McCall model with separation lecture:

### Parameter naming
- Changed utility function parameter from σ (sigma) to γ (gamma) for better consistency with standard CRRA utility notation
- Moved the default value (γ = 2.0) from the utility function to the Model class definition

### Code improvements
- Simplified model parameter unpacking to use direct tuple unpacking: `α, β, γ, c, w, q = model`
- Updated all functions (`compute_v_e`, `update_h`, `solve_model`) to properly pass the γ parameter

### Dynamic figure generation with glue
- Added `myst_nb` glue functionality to replace static PNG figures
- Added `glue()` calls in exercise solution code to capture figures
- Converted all `{figure}` directives to `{glue:figure}` directives
- This eliminates the need for the three static PNG files and ensures figures are always in sync with the code

These changes improve code maintainability and follow better practices for Jupyter Book documentation.

jstac:
@mmcky @HumphreyYang 

Me and my new best buddy Claude used glue to get rid of some static assets -- png files.  Deleting files and dependencies is deeply satisfying.  No more worrying about whether the png files are out of sync.

PR #720, opened 2025-11-20
MAINT: upgrade quantecon-book-theme==v0.11.0

mmcky:
This PR tests the latest version of quantecon-book-theme (v0.11.0).

Changes:
- Updated quantecon-book-theme from 0.10.1 to 0.11.0 in environment.yml

Release notes: https://github.com/QuantEcon/quantecon-book-theme/releases/tag/v0.11.0

This will trigger CI builds to verify compatibility with the new theme version.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - pip
   - pip:
     - jupyter-book==1.0.4post1
-    - quantecon-book-theme==0.10.1
+    - quantecon-book-theme==0.11.0
     - sphinx-tojupyter==0.4.0
     - sphinxext-rediraffe==0.2.7
     - sphinx-exercise==1.2.0


mmcky:
@jstac using `linear_algebra` as a test ground

https://pr-720--sunny-cactus-210e3e.netlify.app/linear_algebra#linear-independence

It looks like emphasis is used a lot instead of definition:

For example

In particular, a collection of vectors $A := \{a_1, \ldots, a_k\}$ in $\mathbb R ^n$ is said to be

* *linearly dependent* if some strict subset of $A$ has the same span as $A$.
* *linearly independent* if it is not linearly dependent.

--- Sonnet 4.5. 

This is a definition. The text linearly dependent appears in a bulleted list where the concept is being formally defined for the first time. According to the QuantEcon style guide, this should use bold formatting instead of italic.

The correct formatting should be:

**linearly dependent** (bold for definition)
Similarly, on line 403, *linearly independent* should also be **linearly independent** since it's also being defined in that same context.

--- seems to do a pretty good job of identifying different cases.

jstac:
> It looks like emphasis is used a lot instead of definition:

Thanks @mmcky 

Yes, we had no policy for this issue for a long time, so it's a complete mess.  

(Hence the suggestion to unify --- but I can see why you rejected that idea.)

mmcky:
- [x] inconsistencies will be fixed by https://github.com/QuantEcon/lecture-python.myst/pull/723

PR #734, opened 2025-11-23
Refactor cross-section simulation: reverse loop structure for better performance

jstac:
## Summary

This PR refactors the cross-sectional agent simulation in both McCall model lectures to use a more efficient and modular loop structure.

### Key Changes

**Previous approach:** Loop over time steps, vectorize over all agents at each step
**New approach:** Vectorize over agents, with each agent looping over time internally

### Implementation Details

1. **Added `sim_agent()` function** - Uses `lax.fori_loop` to simulate a single agent forward T time steps with `fold_in` for key generation
2. **Added `sim_agents_vmap`** - Vectorizes `sim_agent` across multiple agents using `jax.vmap`
3. **Updated `simulate_cross_section()`** - Now generates n_agents keys and passes each to `sim_agent`
4. **Updated `plot_cross_sectional_unemployment()`** - Uses `sim_agents_vmap` directly
5. **Added explanatory text** - Clarifies the difference between `simulate_employment_path()` (records full history for visualization) and `sim_agent()` (returns only final state for efficiency)

### Performance

Testing with 50,000 agents over 200 periods showed:
- **With `@jax.jit` on `sim_agent`:** New approach is ~1.07x faster and has lower variance
- The JIT compilation of the intermediate function before vmapping improves performance

### Files Modified

- `mccall_model_with_sep_markov.md` (discrete wage case)
- `mccall_fitted_vfi.md` (continuous wage case)

Both files now use the same efficient pattern, with the only difference being continuous vs discrete wages.

### Testing

Both notebooks have been converted to Python and run successfully, producing expected results that match the ergodic theorem (time-average ≈ cross-sectional average).

### DIFF
--- a/lectures/mccall_fitted_vfi.md
+++ b/lectures/mccall_fitted_vfi.md
@@ -611,55 +611,50 @@ When employed, the agent faces job separation with probability $\alpha$ each per
 
 Now let's simulate many agents simultaneously to examine the cross-sectional unemployment rate.
 
-We first create a vectorized version of `update_agent` to efficiently update all agents in parallel:
+To do this efficiently, we need a different approach than `simulate_employment_path` defined above.
+
+The key differences are:
+
+- `simulate_employment_path` records the entire history (all T periods) for a single agent, which is useful for visualization but memory-intensive
+- The new function `sim_agent` below only tracks and returns the final state, which is all we need for cross-sectional statistics
+- `sim_agent` uses `lax.fori_loop` instead of a Python loop, making it JIT-compilable and suitable for vectorization across many agents
+
+We first define a function that simulates a single agent forward T time steps:
 
 ```{code-cell} ipython3
-# Create vectorized version of update_agent
-update_agents_vmap = jax.vmap(
-    update_agent, in_axes=(0, 0, 0, None, None)
-)
-```
+@jax.jit
+def sim_agent(key, initial_status, initial_wage, model, w_bar, T):
+    """
+    Simulate a single agent forward T time steps using lax.fori_loop.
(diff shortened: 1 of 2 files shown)


jstac:
## Additional Context

This refactoring was motivated by exploring different strategies for simulating cross-sections in JAX:

### Architecture Comparison

The original implementation used:
```python
def update(t, loop_state):
    # For each time step t:
    # 1. Generate keys for all agents
    # 2. Update all agents in parallel (vmap)
    lax.fori_loop(0, T, update, ...)  # outer loop over time
```

The new implementation reverses this:
```python
def sim_agent(key, ...):
    def update(t, state):
        # For a single agent:
        # Update state using fold_in for keys
    lax.fori_loop(0, T, update, ...)  # inner loop over time

sim_agents_vmap(...)  # outer vectorization over agents
```

### Why This Matters

1. **Modularity**: `sim_agent` is a self-contained function that can be tested and reused independently
2. **Conceptual clarity**: Each agent's trajectory is simulated as a unit, which matches the economic interpretation
3. **Performance**: The JIT compiler can optimize `sim_agent` before it gets vmapped, leading to better performance
4. **Flexibility**: Easier to extend (e.g., returning additional statistics per agent)

### Interesting Finding

Removing `@jax.jit` from `sim_agent` made the new approach slower, which goes against the usual advice to avoid JIT on intermediate functions. In this case, JIT-compiling before vmapping helps the compiler optimize the `lax.fori_loop` structure.

Both implementations are correct and produce statistically equivalent results. The new one is cleaner and performs just as well (or slightly better).

PR #763, opened 2025-12-11
[harrison_kreps] Update lecture

xuanguang-li:
This pull request updates the Harrison-Kreps model lecture to use JAX and improves code consistency and formatting throughout the document.

* Replaced `numpy` (`np`) and `scipy.linalg` (`la`) imports and usages with `jax.numpy` (`jnp`) equivalents throughout code cells, but kept `for` and `while` loops for simplicity. 

* Formatting and terminology improvements: standardized section and subsection titles, and improved consistency in emphasis (e.g., italicizing instead of bolding, fixing typos in comments and variable names).


jstac:
Thanks @xuanguang-li , much appreciated.

I've reviewed the PR.

Actually I don't think switching to JAX is justified in this lecture.  

In the computations of the stationary distribution at the start, we will be using NumPy internally, since that's all qe.MarkovChain can work with.

In the later computation, with the for loop, JAX is probably slower, since the array computations are very low dimensional and the main task is iterative.

I think it might be best to close this PR.

@HumphreyYang or @mmcky, what do you think?  If we do close this, let's tick it off or remove it from the meta issue, so we know not to try again.

HumphreyYang:
Many thanks @jstac, I agree!

@xuanguang-li also has some other valuable edits in this PR so I can  revert the JAX edits and keep others.

xuanguang-li:
Thanks, @jstac and @HumphreyYang, for your comment. 

I agree with your points. JAX usage is minimal in this lecture. Since the equilibrium computation isn't computationally heavy, I kept the `for` and `while` loops unchanged, so JAX is mainly for doing matrix algebra.

jstac:
Thanks @xuanguang-li and @HumphreyYang 

@HumphreyYang , it would be great if you could revert the jax edits.

There are lots of edge cases and in this lecture I think we are making the right decision.

jstac:

**Netlify preview:** https://pr-763--sunny-cactus-210e3e.netlify.app/ is currently **live** — but it's a lucky survivor. It was built 2026-06-05, before the breakages below, and simply hasn't been rebuilt since. **Any rebuild will fail** until the branch is updated (see below).

**1. This branch is stale — 39 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~39 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-763--sunny-cactus-210e3e.netlify.app/harrison_kreps.html
```

---
*This PR touches: `harrison_kreps.md`. Last CI build: `success@2026-06-05`. Branch: 39 commits behind `main` as of 2026-07-08.*

jstac:
Following up on the decision from May — @jstac concluded that JAX isn't justified here (the stationary distributions go through `qe.MarkovChain`, which is NumPy internally, and the remaining work is a low-dimensional iterative loop where JAX would likely be slower), @HumphreyYang agreed and offered to revert the JAX edits while keeping the rest, and @xuanguang-li agreed too. That last step never got finished, so I've done it now.

**Where things stood.** Humphrey's commits on 2026-06-05 had already removed the JAX changes — the branch head contained no `jnp` at all. What remained was a mix: the prose improvements, plus a set of reversions the branch was still carrying against `main` — the jupytext `format_version`/`jupytext_version` fields dropped, `{code-cell} ipython3` back to `{code-cell} ipython`, `:tags: [hide-output]` back to the older `---`-delimited form, and the trailing newline removed. Those came from the branch being 51 commits behind `main`, not from anything anyone did deliberately.

**What I did.** Merged current `main`, resolved `harrison_kreps.md` in its favour to clear those reversions, then reapplied the prose work on top:

- eleven section and subsection headings to sentence case, per the style guide
- typos: "themsevles" → "themselves", "agaub" → "agent"
- bold → italic in the three places where the text is emphasis rather than a definition, which is the convention in the style guide
- the missing `## Exercises` heading before the exercise block

The PR is now +19/−17 in one file.

**One thing worth a second look, @jstac.** The branch changed the $\check p(s)$ minimum equation from

$$P_a(s,0)\,\check p(0) + P_a(s,1)\,(1 + \check p(1))$$

to `P_a(s,1)` on the first term. I think that's the wrong direction: the first term is the probability of moving to state 0 times $\check p(0)$, and equation `hakr2` above has it as `P_a(s,0)`. What the branch was reacting to is real, though — the marginal-investor condition a few lines below `hakr2` does read

$$P_a(s,1)\,\bar p(0) + P_a(s,1)\,(1 + \bar p(1)) < P_b(s,1)\,\bar p(0) + P_b(s,1)\,(1 + \bar p(1))$$

with `P_a(s,1)` and `P_b(s,1)` on the first term, which is inconsistent with `hakr2` directly above it. So I've left the minimum equation as it was and fixed that condition to `P_a(s,0)` / `P_b(s,0)` instead. This is the one place I've changed mathematics rather than prose, so it's worth confirming before merge.

Thanks @xuanguang-li — the heading, typo and emphasis fixes are all worth having, and the equation inconsistency was a good catch even though the fix went the other way.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/239
**Files synced** (1):
- `lectures/harrison_kreps.md`

PR #764, opened 2025-12-14
[rational_expectations] Update content and styling

xuanguang-li:
This pull request makes a series of stylistic and clarity improvements to the `lectures/rational_expectations.md` file. There are no code changes, as the control problem is solved using `quantecon.LQ`, for which `numpy` works well.

Key changes:

**Mathematical and technical clarifications:**
* Clarified the meaning of $Y_{t+1}$ in different places.
* Updated the Euler equation to use the belief function $H$ consistently, and added an equivalent form for clarity.
* Added an explanation for the failure of contractivity using Blackwell's condition, improving conceptual understanding.
* Clarified and corrected the law of motion for the case of a unit measure of firms in the first exercise.
 
**Formatting and reference improvements:**
* Converted links to internal documentation references
* Added a proper definition block.
* Changed all section and subsection headings to sentence case


jstac:

**Netlify preview:** https://pr-764--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 94 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~94 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-764--sunny-cactus-210e3e.netlify.app/rational_expectations.html
```

---
*This PR touches: `rational_expectations.md`. Last CI build: `success@2025-12-14`. Branch: 94 commits behind `main` as of 2026-07-08.*

jstac:
Thanks @xuanguang-li — this one is in good shape, and I've merged current `main` and made a few fixes on top rather than leave it sitting.

Worth saying first: declining to convert this lecture to JAX was the right call, and the reasoning in the PR description is exactly right. The control problem goes through `quantecon.LQ`, which is NumPy internally, so a conversion would have moved the array constructors and left the computation where it was.

**One substantive fix.** Changing the Euler equation {eq}`ree_comp7` to carry $H(Y_t)$ rather than $Y_{t+1}$ is a good clarification — it makes the belief-dependence explicit and matches your new note that $Y_{t+1}$ means perceived output at that stage. But it has a knock-on effect two sections later. **Key insight** said:

> Return to equation {eq}`ree_comp7` and set $y_t = Y_t$ for all $t$. A small amount of algebra will convince you that when $y_t=Y_t$, equations {eq}`comp16` and {eq}`ree_comp7` are identical.

With $Y_{t+1}$ in the Euler equation that really was algebra alone. With $H(Y_t)$, matching {eq}`comp16` also needs $H(Y_t) = Y_{t+1}$, which is the equilibrium fixed-point condition. So I've stated it:

> In a rational expectations equilibrium the perceived and actual laws of motion agree, so $H(Y_t) = Y_{t+1}$.
>
> Imposing this alongside $y_t = Y_t$, a small amount of algebra will convince you that equations {eq}`comp16` and {eq}`ree_comp7` are identical.

**A likely build breaker.** In the new derivation for the unit measure of firms, the second integral ended `dω` with a literal Unicode omega (U+03C9) where the neighbouring one correctly used `d\omega`. MathJax might have coped; the LaTeX builder almost certainly would not. That block was also missing blank lines on both sides, so the `$$` ran straight into the preceding sentence and into `{solution-end}` — enough to break the rendering on its own. Both fixed. The mathematics itself checks out against the $n$-firm result just above it at $n = 1$.

**The Blackwell paragraph** is a genuinely useful addition and I've kept the argument, with some tidying: `intuively` → `intuitively`, `monotoncity` → `monotonicity`, `the Blackwell's condition` → `Blackwell's condition`, and `$y_{t+1} = h(Y_t, Y_t)$` → `$Y_{t+1}$`, since $h(Y_t, Y_t)$ is the aggregate law of motion rather than an individual firm's output. I also split it into one-sentence paragraphs to match the lecture, and made the conclusion explicit — that $\Phi$ *reverses* the ordering of beliefs, which is why monotonicity fails.

Minor: added the missing space in `{doc}`... <lqcontrol>``, matching the other two roles in the lecture.

The branch was 57 commits behind `main`, which is why the preview was 404ing; that's merged now and the PR is marked ready. The lecture executes clean.

jstac:
Thanks @xuanguang-li , comments from Claude above.  This should be good to merge when green --- I'll do so if it looks good.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/243
**Files synced** (1):
- `lectures/rational_expectations.md`

PR #770, opened 2025-12-22
Fix: Remove @jax.jit decorator from excess demand function to resolve shape errors

mmcky:
## Issue

The `e()` excess demand function had a `@jax.jit` decorator that caused shape mismatch errors when called with different dimensional inputs throughout the notebook:
- First called with 2D inputs (A is 2×2, p is length 2)
- Later called with 3D inputs (A is 3×3, p is length 3)

This resulted in: `TypeError: dot_general requires contracting dimensions to have the same shape, got (3,) and (3000,)`

## Solution

Removed the `@jax.jit` decorator from the `e()` function definition at line 556 in newton_method.md.

The `newton()` solver function still uses `@jax.jit` internally for its update function `q()`, so performance is maintained while allowing JAX to handle varying input dimensions flexibly.

## Testing

This fix allows the notebook to execute successfully with the upgraded anaconda environment (2025.12).

## Related

- Discovered while testing PR #768 (anaconda upgrade) with build cache disabled for full execution validation
- Based on upgrade-build-software-2025 branch

### DIFF
--- a/lectures/newton_method.md
+++ b/lectures/newton_method.md
@@ -552,7 +552,6 @@ $$
 The function below calculates the excess demand for given parameters
 
 ```{code-cell} ipython3
-@jax.jit
 def e(p, A, b, c):
     return jnp.exp(-A @ p) + c - b * jnp.sqrt(p)
 ```


mmcky:
Closing as looked more closely at the issue and incorporated a fix in #768

PR #772, opened 2025-12-26
[exchangeable] Simplify code and bring lecture into style-guide compliance

kp992:
Update the lecture to use JAX and removes numba related code. Also fixes minor typos and code styling issues.


jstac:
@longye-tian , would you be willing to review this PR?

@mmcky The deploy link seems broken.  Do they go stale?  after how long?  Should @longye-tian trigger a new build with an empty commit?

jstac:
@longye-tian or any other reviewer: Does adding JAX improve this lecture?  Please also make this value judgement.

longye-tian (commented):
Hi @kp992 @jstac,

I think the JAX conversion is natural in the appendix simulation section, especially around the replacement of the old nested-loop simulation. This seems like a good fit because the simulation has two parts: recursion over time and repetition across independent paths.

It might be helpful to say this explicitly in the text before the code. Currently the lecture says:

> To proceed, we create some Python code.

Perhaps add something like:

> The simulation has two dimensions: recursion over time and repetition across independent paths. In JAX, we use `lax.scan` for the time recursion and `vmap` to apply the path simulator across many independent key sequences.

For the `learning_example` block, I’m less sure that the JAX conversion adds much. This part changes the beta-density helper from the old Numba/vectorized version and wraps several scalar helpers with `jax.jit`, but the surrounding computation is still mostly SciPy root finding, SciPy quadrature, NumPy grids, and Matplotlib plotting. So JAX transformations do not seem central to this part of the lecture. I can see the value of removing Numba and using JAX for consistency, but plain NumPy/SciPy may be clearer here.

A more useful exposition improvement might be to split this page-long helper function into the three conceptual pieces shown in the output figure:

1. likelihood ratio plot
2. density / probability-region plot
3. posterior-dynamics arrows

That would let the text introduce and discuss each panel one by one, rather than asking readers to parse a long plotting helper before seeing the three ideas it creates.

**A few small code-level updates could also make the JAX appendix clearer:**

- Instead of `random_seed = int(a * b + T + N)`, consider adding an explicit `seed` or `key` argument. That would make the source of randomness clearer.
- The comment `# Generate all random keys upfront` appears inside `simulate_path`, but the keys are actually generated in `simulate`; perhaps change it to something like `# Use one random key for each date`.

**A few small remaining typos I noticed while reading:**

- “less that one” → “less than one”
- “Bayes’ Law make $\pi$ decrease” → “Bayes’ Law makes $\pi$ decrease”
- “The above graphs shows” → “The above graph shows”
- `expected_rario` → `expected_ratio`

Best,
Longye

jstac:

**Netlify preview:** https://pr-772--sunny-cactus-210e3e.netlify.app/ currently returns **404**.

**1. This branch is stale — 93 commits behind `main`.** A preview build compiles the *whole site from this branch*. This branch's `lectures/house_auction.md` still has unpinned `!pip install prettytable`, which now breaks on a `wcwidth` incompatibility. `main` fixed this on 2026-06-28 by pinning `prettytable<3.18` (#939). **This alone fails any rebuild of this branch until it's updated to `main`.**

Update this branch to `main` (merge or rebase — pulls in #939 plus ~93 other commits), then let CI rebuild. On today's latest libraries the site builds clean, so the preview should return. `house_auction` is the *known* blocker; updating also picks up other since-merged fixes — rebuild and address any remaining per-lecture failures. Verify with:

```
curl -sI https://pr-772--sunny-cactus-210e3e.netlify.app/exchangeable.html
```

---
*This PR touches: `exchangeable.md`. Last CI build: `failure@2026-07-07`. Branch: 93 commits behind `main` as of 2026-07-08.*

kp992:
Thanks @longye-tian. I will review your comment and make the changes.

jstac:
Thanks @kp992 for the work here, and @longye-tian for a careful review — the review is what turned this around, so let me explain where it landed and why.

## Summary

I've changed direction on this PR: rather than extend the JAX conversion, we **drop JAX entirely** and keep the lecture in NumPy/SciPy. Numba goes too. I've also acted on Longye's structural suggestion and taken the opportunity to fix some pre-existing style-guide violations. I've pushed this to the branch and retitled the PR accordingly.

## Why not JAX

Longye's instinct about `learning_example` was right, and the [JAX style guide](https://manual.quantecon.org/styleguide/jax.html) names this case explicitly under *when not to use JAX*:

> **Converting only the plots.** Figure generation is never the bottleneck.

and

> Substituting `jnp` for `np` is *not* converting a lecture to JAX. If the numerical work still runs through SciPy, `scipy.stats`, or NumPy-based QuantEcon.py classes, then the arrays are merely being shuttled between host and device at every call. The lecture gets JAX's constraints and none of its benefits.

That is exactly the situation in `learning_example` and `expected_ratio`: the actual computation is `scipy.optimize.root_scalar` and `scipy.integrate.quad` over 100-point plotting grids. Measured locally, results were identical but the jitted version of that section took **0.245s against 0.0004s** for plain NumPy — roughly 600× slower, because `quad` and `root_scalar` call the jitted function scalar-by-scalar and pay dispatch overhead on every one.

That left only the appendix simulation as a genuine JAX candidate, and that turned out to have a much better answer than either `lax.scan` or a Numba loop — see below. With no JAX work left, `jax.config.update("jax_enable_x64", True)` went too. (For the record I did test whether it was needed: with the same draws, float32 and float64 belief paths differ by at most 1.5e-3, which is invisible among 1000 plotted paths.)

## What replaced it

**1. The simulation is now a cumulative product.** Writing Bayes' Law in odds form,

$$\frac{\pi_{t+1}}{1-\pi_{t+1}} = l(w_{t+1})\,\frac{\pi_t}{1-\pi_t} \quad\Longrightarrow\quad \frac{\pi_t}{1-\pi_t} = \frac{\pi_{-1}}{1-\pi_{-1}}\prod_{s=0}^{t} l(w_s)$$

so the whole ensemble of belief paths is a `np.cumprod` of likelihood ratios — no `lax.scan`, no `vmap`, no Python loop. `simulate` is now four lines. Both equations are in the lecture, so the code follows visibly from the mathematics rather than looking like a trick, and it earns the sentence about likelihood ratio processes that was already in the text but previously unexplained. Verified against the sequential recursion: agreement to ~1e-12.

**2. `learning_example` is split, as Longye suggested.** There is now a `create_model` that builds `f`, `g` and the two roots of `l(w) = 1`, plus `plot_likelihood_ratio`, `plot_densities` and `plot_belief_dynamics` — one per panel. The text introduces and discusses each graph in turn instead of asking the reader to parse a page-long helper before seeing the three ideas it produces.

**3. Random number generation** now uses `np.random.default_rng` with an explicit generator passed in, per the [style guide](https://manual.quantecon.org/styleguide/code.html#numpy-random-number-generation).

## Style-guide fixes

Some of these predate this PR, but they were worth fixing while the file was open:

- Four instances of ``{doc}`this lecture <...>` `` replaced with the auto-title form (``{doc}`odu` `` etc.) — see [Using lecture titles in `{doc}` references](https://manual.quantecon.org/styleguide/doclinks.html#using-lecture-titles-in-doc-references). This file had more of these than any other lecture in the repo.
- The hard-coded `python-advanced.quantecon.org` URL replaced with the intersphinx reference ``{doc}`advanced:additive_functionals` ``.
- Seven section headings lower-cased — full capitalization is for the lecture title only.
- kp992's typo fixes retained, plus two more found on `main`: "If a sequence **is** random variables is IID" and "something to learn about **about** the future".

## Verification

All twelve figures were rendered and compared against `main`. The three original panels reproduce **exactly** — same curves, same shaded regions, same printed areas — so the hard-coded numbers in the prose (0.524, 0.476, 0.816) still hold. The simulation figures match in shape and magnitude, differing only through the new generator, as expected.

Runtime for the whole lecture:

| version | time |
|---|---|
| this branch | **2.6s** |
| `main` (Numba) | 3.4s |
| previous state of this PR (JAX) | 7.8s |

One footnote for anyone reading the diff: I first wrote the Beta density as `scipy.stats.beta.pdf`, which reads more cleanly, and it took the lecture from 3.4s to **47s**. `beta.pdf` costs ~50µs per call against ~0.3µs for the explicit `math.gamma` formula, and `quad` calls it scalar-by-scalar. Worth remembering generally — `scipy.stats` `.pdf` methods are a poor fit inside quadrature.

@kp992, apologies that this reworks your PR fairly heavily — the JAX question was one I asked and it needed a maintainer answer, which it didn't get until now. The typo fixes and the structural cleanup you started are all still here. @longye-tian, thank you — both of your substantive points are now addressed, one by removal rather than by revision.

jstac:
@kp992 @longye-tian The above report is written by Claude.  

It's getting much easier to determine what should go to JAX and what should stay -- just point Claude at the style guide (manual) and ask it to assess the value of the conversion based on the benchmarks in that style guide.

In this case Claude finds that there's no real value in shifting to JAX so I'm reverting to NumPy.

I'm following @longye-tian 's suggestion for improving readability and also taking the opportunity to bring the rest of the lecture in line with the style guide.

Merging this now (CC @mmcky).

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/233
**Files synced** (1):
- `lectures/exchangeable.md`

PR #822, opened 2026-03-07
Add Reinforcement Learning section

jstac:
## Summary

- Adds a new **Reinforcement Learning** section to the book
- Includes a new lecture `inventory_q.md`: inventory management solved via value function iteration, then learned via tabular Q-learning
- Moves `mccall_q.md` from the Search section into the new RL section

## Details

The new lecture covers:
- An inventory management model with stochastic demand and fixed ordering costs
- Exact solution via VFI (Bellman operator, greedy policy, simulation)
- Derivation of the Q-factor Bellman equation from the standard Bellman equation
- Q-learning with epsilon-greedy exploration on a single continuous trajectory
- Comparison of learned vs optimal value functions and policies
- Visualization of policy evolution during training

### DIFF
--- a/lectures/_config.yml
+++ b/lectures/_config.yml
@@ -107,8 +107,8 @@ sphinx:
     mathjax3_config:
       tex:
         macros:
-          "argmax" : "arg\\,max"
-          "argmin" : "arg\\,min"
+          "argmax" : ["\\operatorname*{argmax}", 0]
+          "argmin" : ["\\operatorname*{argmin}", 0]
     mathjax_path: https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js
     # Local Redirects
     rediraffe_redirects:
--- a/lectures/inventory_q.md
+++ b/lectures/inventory_q.md
@@ -654,7 +654,7 @@ We extract the value function and policy from the final Q-table via
 $$
     v_Q(x) = \max_{a \in \Gamma(x)} q(x, a),
     \qquad
-    \sigma_Q(x) = \arg\max_{a \in \Gamma(x)} q(x, a)
+    \sigma_Q(x) = \argmax_{a \in \Gamma(x)} q(x, a)
 $$
 
 and compare them against $v^*$ and $\sigma^*$ from VFI.


jstac:
@mmcky , please take a look when you have time

jstac:
@mmcky , please take a look when you have time.  Thanks.

(I notice that the branch is not deployed...?)

jstac:
Something a big strange happening with the code cells:

image

(Sorry @mmcky , I realize I said I wouldn't be needing to edit lectures but I've decided to do my teaching next month at CMU directly from the website, so I'll probably do some editing of this series.  Also, this new section 'Reinforcement Learning' should be shifted onto the DP lectures when you set that up.)

mmcky:
thanks @jstac for reporting the issue from the preview. 

This lecture looks really great. I'm running out the door now, but will review this soon.

mmcky:
## Notation Issues

### 1. Transition function argument order mismatch

The transition dynamics are defined as:

$$X_{t+1} = h(X_t, D_{t+1}, A_t)$$

but the helper is defined as:

$$h(x, a, d) := (x - d) \vee 0 + a$$

The second argument of $h$ in the first equation is $D_{t+1}$ (demand), but in the definition it is $a$ (the order quantity). The arguments are swapped. It should be either:

- $X_{t+1} = h(X_t, A_t, D_{t+1})$ with $h(x, a, d)$, **or**
- $h(x, d, a) := (x - d) \vee 0 + a$

The code itself is correct — this is a notation-only issue.

### 2. Reward function notation: `r` vs `π`

The profit function is defined as $\pi(X_t, A_t, D_{t+1})$, but the Bellman equation and Q-factor definitions use $r(x, a, D)$ without ever defining $r$. These should be unified — either use $r$ throughout or add a line like $r \equiv \pi$.

jstac:
thanks @mmcky , working on these now.

jstac:
## Review and fixes for `inventory_q.md`

I did a careful read-through of the lecture, looking for mathematical errors, notational issues, and clarity for readers who know dynamic programming but not reinforcement learning. Here's a summary of all changes made.

---

### Mathematical errors fixed

**1. Argument order in the transition function `h`**

The transition was written as

$$X_{t+1} = h(X_t, D_{t+1}, A_t)$$

but `h` was defined as $h(x, a, d) := (x - d) \vee 0 + a$, which treats the second argument as the action (added to stock) and the third as demand (subtracted). Under the original calling convention, the arguments were swapped — demand was passed as the second argument and action as the third.

The rest of the lecture (Bellman equation, Q-factor equation) consistently uses the `(x, a, d)` convention, so the fix was to change the call to $h(X_t, A_t, D_{t+1})$.

**2. Notation clash: `r` used for both interest rate and reward function**

The interest rate was denoted $r > 0$ (in the definition $\beta = 1/(1+r)$), while the reward function in the Bellman equation and Q-factor equations was also denoted $r(x, a, d)$. Since per-period profits were already denoted $\pi_t$ earlier in the lecture, we renamed the reward function to $\pi(x, a, d)$ throughout, which is consistent and avoids the clash.

**3. Action space typography**

The state space was written in mathsf ($\mathsf{X}$) but the action space definition used plain math ($A := X$). Fixed to $\mathsf{A} := \mathsf{X}$ for consistency.

**4. Inconsistent notation in the modified update rule**

The new section contrasting Q-learning with policy evaluation used $\pi_{t+1}$ in the modified update rule, while the original Q-learning update used $R_{t+1}$. The subscript $\pi_{t+1}$ is ambiguous (it could mean profit at time $t+1$, which would be $\pi(X_{t+1}, A_{t+1}, D_{t+2})$). Changed to $R_{t+1}$ for consistency.

---

### Prose improvements

**5. Timing language**

The original text said orders "arrive at the start of period $t+1$, after the firm caters to current demand $D_{t+1}$." The phrase "current demand $D_{t+1}$" is confusing since $D_{t+1}$ is next-period demand. Changed to "after demand $D_{t+1}$ is realized and served."

**6. Major rewrite: Q-table and behavior policy section**

This was the most substantial change. The original text stated:

> "In principle, the manager could choose $a$ completely at random and the Q-table would still converge to $q^*$, provided every $(x, a)$ pair is visited infinitely often."

This claim is correct *in the RL sense*, but for a reader coming from a mathematics/DP background, it reads as false. Here's why: the update rule contains a $\max_{a'}$, and the natural mathematical reading is that this $\max$ determines the next action. If you replace the $\max$ with a fixed policy $\sigma$, the Q-table converges to $q^\sigma$ (the Q-function of $\sigma$), not $q^*$. So the claim that "you could choose randomly and still converge" appears to contradict the update rule itself.

The resolution is that the $\max$ in the update target computes a *scalar value* (an estimate of the value of being in the next state under optimal continuation), not an action prescription. Which action the manager actually takes next is a separate decision — it only affects which $(x, a)$ entries get visited, not the convergence target.

The rewritten section now:
- Explicitly states that the $\max$ computes a value, not an action
- Contrasts Q-learning (optimality operator, converges to $q^*$) with the modified update under a fixed policy $\sigma$ (evaluation operator, converges to $q^\sigma$)
- Introduces the **behavior policy** as the separate rule governing action selection
- Introduces **on-policy vs. off-policy** terminology, explaining that Q-learning is off-policy because its convergence target does not depend on the behavior policy
- States the convergence conditions precisely

**7. Code comments improved**

The original comment in `q_learning_kernel` read:
```python
# === Find best action at x_next (used for both update and next step) ===
```

This conflates the two distinct uses of the max computation, which is exactly what the new text works hard to separate. The code computes both `best_next` (the max value, a scalar) and `a_next` (the argmax action). These serve different purposes:
- `best_next` enters the **update target** (line: `q[x, a] = (1 - α) * q[x, a] + α * (reward + β * best_next)`)
- `a_next` feeds the **behavior policy** (line: `a = a_next`, subject to ε-perturbation)

The comments now clearly label each block:
```python
# === Max over next state (scalar value for update target) ===
# Also record the argmax action for use by the behavior policy.
...
# === Q-learning update (uses best_next, the max value) ===
...
# === Behavior policy: ε-greedy (uses a_next, the argmax action) ===
```

jstac:
@mmcky , this is nearly good to go.

One issue is that \argmax is not rendering properly (it should be rm script, like \max).  When you have time, please fix that issue and then make this live.

mmcky:
Fixed `\\argmax` rendering across lectures. Updated the global MathJax macros in `_config.yml` to use `\\operatorname*{argmax}` and `\\operatorname*{argmin}`, so subscripts render directly below in display mode (matching `\\max`). This is a global fix — no per-file workarounds needed.

jstac:
Love your work @mmcky , much appreciated.

mmcky:
thanks @jstac -- I'll issue a publish now.

PR #824, opened 2026-03-07
Pin quantecon-book-theme to 0.16.0

mmcky:
Pin quantecon-book-theme back to 0.16.0 due to dark code block background leakage in light mode.

The issue was introduced in v0.17.0 (color scheme system overhaul), not v0.18.0. v0.16.0 was the last published version without this problem. Also removes `sticky_contents: true` from `_config.yml` as it is a v0.17.0 feature.

See: https://github.com/QuantEcon/quantecon-book-theme/issues/372

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - pip
   - pip:
     - jupyter-book>=1.0.4post1,<2.0
-    - quantecon-book-theme==0.17.1
+    - quantecon-book-theme==0.16.0
     - sphinx-tojupyter==0.6.0
     - sphinxext-rediraffe==0.3.0
     - sphinx-exercise==1.2.1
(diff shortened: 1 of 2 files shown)


mmcky:
@jstac @HumphreyYang I am going to revert the theme back to 0.17.1 while I investigate the code cell issue. 

This hasn't been on the `live` site until just 15minutes ago. So this should fix the issue.

mmcky:
@HumphreyYang and @jstac this doesn't appear to be related to our theme. More of an ecosystem shift, I am looking into it but it doesn't appear to be a quick fix. 

If I can't resolve it @HumphreyYang do you need the current live updates? I could revert the live site later this evening if a fix isn't presenting itself.

HumphreyYang:
Many thanks @mmcky. I’ve just published a new version of Tom’s lectures, so if it doesn’t affect the presentation in a major way, perhaps we should keep it as is!

mmcky:
@HumphreyYang thanks. Code cells are visible but hard to read. I'll keep investigating first.

mmcky:
@jstac @HumphreyYang I think I have found the culprit - a downstream commit hidden in mystnb. 

Hope to re-publish after https://github.com/QuantEcon/lecture-python.myst/pull/825 verifies.

PR #867, opened 2026-05-05
FIX: Define missing discount factor to resolve NameError in MPE notebook

HG-Cheng:
Description:
In the Markov Perfect Equilibrium lecture, the cell calculating the LQ stationary values throws a NameError during notebook execution. This occurs because the discount factor (beta = 0.96) is defined inside the external script duopoly_mpe.py. Since the :load: directive only renders the script statically without executing it in the active Jupyter kernel, beta remains undefined in the notebook namespace.

This PR explicitly defines beta = 0.96 in the execution cell to ensure the qe.LQ model initializes correctly and the notebook builds successfully.


jstac:
Many thanks @HG-Cheng , much appreciated!

PR #874, opened 2026-05-25
[lln_clt.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `lln_clt.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

The single legacy call `np.random.randint` inside `gen_x_draws` has been replaced with `rng.integers` from a `np.random.default_rng()` generator.

## Details

`rng` is introduced as an explicit required argument to `gen_x_draws(k, rng)`, and `rng = np.random.default_rng()` is defined in the same cell just before the call site. Since `rng` is always available at the point of use, no `rng=None` fallback was added.

This differs from the approach taken in a related PR ([lecture-python-intro#741](https://github.com/QuantEcon/lecture-python-intro/pull/741/changes)), where `rng` was created once at the top of the lecture and reused across multiple cells — so functions there could be called from contexts where `rng` was not always explicitly passed, making the `if rng is None` fallback reasonable. In this lecture, `rng` is created in the same cell as the function definition and its call site, so the fallback seemed unnecessary.

I'd welcome your thoughts on whether this difference in approach is appropriate, or whether a consistent `rng=None` pattern should be preferred across lectures.

### DIFF
--- a/lectures/lln_clt.md
+++ b/lectures/lln_clt.md
@@ -444,7 +444,7 @@ $Y_5$
 ```{code-cell} python3
 beta_dist = beta(2, 2)
 
-def gen_x_draws(k):
+def gen_x_draws(k, rng):
     """
     Returns a flat array containing k independent draws from the
     distribution of X, the underlying random variable.  This distribution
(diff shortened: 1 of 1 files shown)


jstac:
Thanks @Chihiro2000GitHub

PR #890, opened 2026-06-02
[blackwell_kihlstrom.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `blackwell_kihlstrom.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

## Details

- Replaced `np.random.seed(0)` with `rng = np.random.default_rng(0)` in the imports block.
- Updated `sample_posteriors` to accept `rng=rng` as a default argument and replaced the two internal `np.random.choice(...)` calls with `rng.choice(...)`.
- Call sites for `sample_posteriors` are unchanged — the module-level `rng` is used by default.
- `sequential_update` and `check_martingale_mean` were already using the Generator API and required no changes.
- No fixed seed was introduced beyond what the original `np.random.seed(0)` already provided.
- No Numba-related code in this lecture.


lectures/blackwell_kihlstrom.md
```diff
@@ -627,15 +627,15 @@ mystnb:
     caption: Sampled posterior points on the 2-simplex
     name: fig-blackwell-simplex-clouds
 ---
-def sample_posteriors(μ_matrix, prior, n_draws=3000):
+def sample_posteriors(μ_matrix, prior, n_draws=3000, rng=rng):
```
mmcky: Thanks @Chihiro2000GitHub — this is correct and faithfully preserves the original behaviour (the two `sample_posteriors` calls still draw sequentially from a single stream).

One small consistency point: this lecture already has two sibling functions, `sequential_update` and `check_martingale_mean`, that take a `seed` and build the generator locally (`rng = np.random.default_rng(seed)`). With this change `sample_posteriors` instead relies on a module-level `rng` injected via a `rng=rng` default argument, so the lecture ends up with two different RNG idioms for three very similar functions.

Could we align `sample_posteriors` with its siblings by building the generator in the driver (`plot_simplex_posteriors`) and passing it in? That keeps a single shared stream (so the μ and ν clouds stay independent, as now), removes the module-level global, and lets us drop the `np.random.seed(0)` line at the top. Something like:

```python
def sample_posteriors(μ_matrix, prior, n_draws=3000, rng=None):
    """..."""
    if rng is None:
        rng = np.random.default_rng()
    N, M = μ_matrix.shape
    states = rng.choice(N, size=n_draws, p=prior)
    signals = np.array([rng.choice(M, p=μ_matrix[s]) for s in states])
    ...
```

and then in `plot_simplex_posteriors`:

```python
def plot_simplex_posteriors(μ_matrix, ν_matrix, prior3, n_draws=3000, seed=0):
    rng = np.random.default_rng(seed)
    posts_μ = sample_posteriors(μ_matrix, prior3, n_draws, rng)
    posts_ν = sample_posteriors(ν_matrix, prior3, n_draws, rng)
    ...
```

With that, the `rng = np.random.default_rng(0)` line in the imports block can be removed. Non-blocking — happy either way, but it would make the three functions consistent.

mmcky (commented):
Thanks @Chihiro2000GitHub! The migration is correct and behaviour-preserving. Just one optional consistency nit inline about aligning `sample_posteriors` with the seed-based pattern already used by `sequential_update` / `check_martingale_mean` in this lecture. Otherwise good to go. (Worth a quick look at the [Netlify preview](https://pr-890--sunny-cactus-210e3e.netlify.app/blackwell_kihlstrom.html) to confirm the simplex clouds still read well, since the seeded stream changes under `Generator`.)

@HumphreyYang — could you weigh in on the inline suggestion? It's a judgment call about which RNG idiom to standardise on within a lecture (module-level `rng` vs. seed-per-function), so I'd like your read before we ask for changes.

lectures/blackwell_kihlstrom.md
HumphreyYang: Many thanks @mmcky, that makes sense.

I agree that creating the generator locally in `plot_simplex_posteriors` is
cleaner and keeps the RNG pattern consistent with `sequential_update` and
`check_martingale_mean`.

Changing to

```python
posts_μ = sample_posteriors(μ_matrix, prior3, n_draws, rng=rng)
posts_ν = sample_posteriors(ν_matrix, prior3, n_draws, rng=rng)
```

also makes it explicit that both posterior samples are drawn using the same
local generator.

After making this change, we can drop the global

```python
rng = np.random.default_rng(0)
```

because it was only being used as the default value for `rng` in
`sample_posteriors`.

I think it is a strict improvement and should be adopted everywhere else!

Chihiro2000GitHub: @mmcky @HumphreyYang,

Thank you both! I'll go through your comments later, probably this weekend.

Chihiro2000GitHub:
Thanks very much @mmcky and @HumphreyYang 😊

This is a fair point, and I agree it's a strict improvement.
Could you take another look when you have a moment?

mmcky (approved):
thanks @Chihiro2000GitHub I think this is good to go. 

@HumphreyYang do you mind 👍 when you get a chance.

HumphreyYang:
Many thanks @mmcky, it looks great! Should we merge this so that I can add this to the new release?

PR #907, opened 2026-06-16
[kalman_2] Update code and descriptions

### LINKED ISSUE #904: [kalman_2] Some comments for the section 3 and 4
Hi @jstac,

I noticed a few points in section 3 and 4 of `kalman_2` lecture that might be worth clarifying.

### Section 3 A state-space representation

- The displayed observation equation seems to use $G = [g \quad 0]$, so that output depends on human capital but not directly on effort. In the code, however, we currently set

  ```python
  G = np.array([g, 1])
  ```

  Is this meant to be `np.array([g, 0])`, or should the text explain why effort enters the observation equation directly?

- The covariance matrix $\Sigma_0$ uses $\sigma_{h,0}$ and $\sigma_{u,0}$ as entries. It might be helpful to clarify whether these symbols denote variances or standard deviations.

- The simulated `LinearStateSpace` object is created with `Sigma_0=np.zeros((2,2))`, even though the model has just defined a non-degenerate initial covariance matrix $\Sigma_0$. If this is intentional, perhaps we could add a sentence explaining that the simulation fixes a particular worker's initial state, while the firm still begins with uncertain beliefs about that worker.

- There is a small typo: `we first we create` could be changed to `we first create`.

### Section 4 An Innovations Representation

- The title could perhaps use sentence case: `An innovations representation`.

- After introducing the innovation equation, it might help to define $a_t$ explicitly, for example as the innovation / forecast error:

  $$a_t = y_t - G \hat x_t$$

- The code calls

  ```python
  kalman.update(y[t])
  ```

  before recording `x_hat` and `Σ`. Since `kalman.update(y[t])` incorporates the current observation `y[t]` and then advances the filter to the next predictive state, the recorded `x_hat` does not seem to correspond to the pre-observation belief $\hat x_t = \mathbb{E}[x_t \mid y^{t-1}]$ described in the text. If the plots are meant to show the firm's belief before observing current output $y_t$, perhaps the code should record the current prediction first and then call `kalman.update(y[t])`.

- Relatedly, the loop uses `range(1, T)`, which starts the Kalman-filter loop at `t = 1`, so the first simulated observation `y[0]` is not passed to `kalman.update`. If the firm is meant to observe output from time 0 onward, it would be helpful to clarify whether `y[0]` is intentionally skipped, or whether the update loop should start at `t = 0`.

- The code prepends `x[:, 1]` to `x_hat_t`, even though `x` is the simulated true hidden state. Since `x_hat_t` otherwise seems to represent filtered / predicted estimates, it might be clearer to avoid mixing in the true state, or to explain why this is being done.

- The covariance array prepends `worker.Σ_0`, while the state-estimate array prepends the true simulated state. It might be clearer if both initial entries represented the same object, either the firm's prior belief or the realized worker state.

- The text switches between $u_0$ and $u_t$ in the inference labels. Since $u_t = u_0$, this is not mathematically wrong, but using one notation consistently might be clearer for readers.

- The plot labels use $E[y_t \mid y^{t-1}]$, while the text says “we plot $E y_t = G \hat x_t$”. This could perhaps be written more explicitly as $\mathbb{E}[y_t \mid y^{t-1}] = G \hat x_t$.

- The plots use embedded titles via `set_title`, which may be worth checking against the figure style rule `qe-fig-003`.

Would these updates make sense? I’d be happy to put together a PR if they look good to you.

Best,
Longye

HumphreyYang:
This PR closes #904 and makes a few improvements beyond it.

Hi @longye-tian, this PR resolves #904, and thanks for raising those issues!

### DIFF
--- a/lectures/kalman_2.md
+++ b/lectures/kalman_2.md
@@ -29,7 +29,7 @@ kernelspec:
 :depth: 2
 ```
 
-In this QuantEcon lecture {doc}`kalman`, we used
+In {doc}`kalman`, we used
 a Kalman filter to estimate  locations of a rocket. 
 
 In this lecture,  we'll use the Kalman filter to 
--- a/lectures/kalman_2.md
+++ b/lectures/kalman_2.md
@@ -11,7 +11,7 @@ kernelspec:
   name: python3
 ---
 
-(kalman)=
+(kalman_2)=
 ```{raw} jupyter
 <div id="qe-notebook-header" align="right" style="text-align:right;">
         <a href="https://quantecon.org/" title="quantecon.org">
@@ -83,10 +83,10 @@ Here
 * $u_t$ is the logarithm of the worker's effort at accumulating human capital at $t$ 
 * $y_t$ is the logarithm of the worker's output at time $t$
 * $\epsilon_{t+1}$ is an IID standard normal shock to human capital
-* $h_0 \sim N(\hat h_0, \sigma_{h,0})$
-* $u_0 \sim N(\hat u_0, \sigma_{u,0})$
+* $h_0 \sim N(\hat h_0, \sigma_{h,0}^2)$
+* $u_0 \sim N(\hat u_0, \sigma_{u,0}^2)$
 
-Parameters of the model are $\alpha, \beta, c, R, g, \hat h_0, \hat u_0, \sigma_{h,0}, \sigma_{u,0}$.
+Parameters of the model are $\alpha, \beta, c, R, g, \hat h_0, \hat u_0, \sigma_{h,0}, \sigma_{u,0}$ where $\sigma_{h,0}$ and $\sigma_{u,0}$ are standard deviations of the firm's initial beliefs about $h_0$ and $u_0$.
 
 At time $0$, a firm has hired the worker.
 
@@ -94,17 +94,17 @@ The worker is permanently attached to the firm and so works for the same  firm a
 
 At the beginning of time $0$, the firm observes neither the worker's innate initial human capital $h_0$ nor its hard-wired permanent effort level $u_0$.
 
-The firm believes that $u_0$ for a particular worker is drawn from a Gaussian probability distribution, and so is  described by $u_0 \sim N(\hat u_0, \sigma_{u,0})$.
+The firm believes that $u_0$ for a particular worker is drawn from a Gaussian probability distribution, and so is  described by $u_0 \sim N(\hat u_0, \sigma_{u,0}^2)$.
 
 The $h_t$ part of a worker's "type" moves over time, while the equation $u_{t+1} = u_t$ implies $u_t = u_0$ for all $t$.
 
 Thus, from the firm's point of view, effort is a fixed, unobserved component of the worker's type that must be inferred from output observations.
 
-At time $t\geq 1$, for a particular worker the  firm  observed  $y^{t-1} = [y_{t-1}, y_{t-2}, \ldots, y_0]$.
+At the beginning of time $t \geq 1$, before setting the wage $w_t$, the firm has observed the history $y^{t-1} = [y_{t-1}, y_{t-2}, \ldots, y_0]$.
 
 The firm does not observe the  worker's "type" $(h_0, u_0)$.
 
-But the firm  does observe the worker's  output $y_t$ at time $t$ and remembers the worker's past outputs $y^{t-1}$.
+After production in period $t$, the firm observes the worker's output $y_t$, which it then uses to update its beliefs going into period $t+1$.
 
 ## A firm's wage-setting policy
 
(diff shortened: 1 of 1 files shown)


HumphreyYang:
Hi @longye-tian,

I think this is ready to be merged. Please let me know if there is anything else you would like to add!

lectures/kalman_2.md
```diff
@@ -204,15 +207,17 @@ y = y.flatten()
...
 
+We set `Sigma_0=np.zeros((2,2))` so that the simulation fixes a particular worker's initial state $(h_0, u_0)$, while the firm still enters period $0$ with the non-degenerate prior beliefs $\hat x_0$ and $\Sigma_0$ that drive its Kalman filter.
+
 Next, to  compute the firm's policy for setting the log wage based on the information it has about the worker,
 we  use the Kalman filter described in this QuantEcon lecture {doc}`kalman`.
 
 In particular, we want to compute all of the objects in an "innovation representation".
 
-## An Innovations Representation
+## An innovations representation
```
longye-tian, line 217: Hi @HumphreyYang 

Maybe here we should use `An innovation representation` or `An innovation's representation`.

What do you think?

Best,
Longye

HumphreyYang: Many thanks @longye-tian! Innovations representation is a standard term in timeseries analysis (I know it is a bit strange!)

(See e.g., https://www.stat.berkeley.edu/~bartlett/courses/153-fall2010/lectures/10.pdf)

lectures/kalman_2.md
```diff
@@ -222,61 +227,77 @@ Let's code that up now.
...
+---
 fig, ax = plt.subplots(1, 2)
 
 ax[0].plot(y_hat_t, label=r'$\mathbb{E}[y_t| y^{t-1}]$')
-ax[0].set_xlabel('Time')
-ax[0].set_ylabel(r'$\mathbb{E}[y_t]$')
-ax[0].set_title(r'$\mathbb{E}[y_t]$ over time')
+ax[0].set_xlabel('time')
+ax[0].set_ylabel(r'$\mathbb{E}[y_t| y^{t-1}]$')
+ax[0].set_title(r'$\mathbb{E}[y_t| y^{t-1}]$ over time')
```
longye-tian, line 283: Hi @HumphreyYang 

According to the [style guide `qe-fig-003` No matplotlib embedded titles](https://github.com/QuantEcon/action-style-guide/blob/df55c87595709f6711b1e8bf4f18341ee6009407/docs/user/rules-reference.md), should we delete the title and descibe those figures inline?

Best,
Longye

lectures/kalman_2.md
```diff
@@ -304,31 +331,32 @@ fig, axs = plt.subplots(1, 3, figsize=(12, 7))
...
     # Evaluate the multivariate normal PDF on the grid
     pdf_values = mvn.pdf(np.dstack((h, u)))
     
     # Create a contour plot for the PDF
     con = axs[i].contour(h, u, pdf_values, cmap='viridis')
     axs[i].clabel(con, inline=1, fontsize=10)
-    axs[i].set_title(f'Time Step {t+1}')
-    axs[i].set_xlabel(r'$h_{{{}}}$'.format(str(t+1)))
-    axs[i].set_ylabel(r'$u_{{{}}}$'.format(str(t+1)))
+    axs[i].set_title(f'Time Step {t}')
```
longye-tian, line 344: same stye guide comment here

longye-tian (commented):
Many thanks @HumphreyYang ! 

They all looks good to me. Just some minor comments see above. 

Best,
Longye

lectures/kalman_2.md
longye-tian: Thanks @HumphreyYang!

I see, shall we use the term `The innovations representation` instead of `An innovations representation` to match your referenced document?

Best,
Longye

HumphreyYang:
Many thanks @longye-tian for the detailed check! 

I left out those two cases intentionally.

Personally, I think it would be a bit unclear if we don’t use subplot titles for figures with multiple subplots. 

For instance, if we remove the `Time Step {t}` title from subplots, it would be a bit hard for readers to tell what the plot is showing at first glance.

Please let me know what you think between those two cases! If this makes sense, I can update the style guide accordingly (CC @mmcky and @jstac)!

------

Without subplot titles:
Screenshot 2026-06-16 at 4 06 17 PM

With subplot titles:
Screenshot 2026-06-16 at 4 07 20 PM

jstac:
Many thanks @HumphreyYang @longye-tian !

@HumphreyYang , i agree, subplots should be the exception --- they often need titles.  This particular case is a good example.  It would be great if you could update the style guide -- at titles to subplots as required, for clarity.

HumphreyYang:
Many thanks @jstac, I will open a PR in the manual repo!

I will merge this one for now. 

Many thanks @longye-tian as well for the detailed review!

lectures/kalman_2.md
HumphreyYang: Sorry @longye-tian that I missed this! I think `An innovations representation` is fine because there are a few candidates we can formulate!

PR #961, opened 2026-07-12
[ar1_turningpts.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `ar1_turningpts.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299). Legacy global random-state calls are replaced with an explicit `np.random.default_rng()` generator.

## Details

- Replaced `np.random.seed(145)` with `rng = np.random.default_rng(145)`, preserving the existing fixed seed for reproducibility.
- Replaced `np.random.normal(...)` with `rng.normal(...)` and `np.random.choice(...)` with `rng.choice(...)`.
- The `rng` generator is defined once in the main text and passed explicitly as an argument to `AR1_simulate`, `plot_initial_path`, `plot_Wecker`, and `plot_extended_Wecker`, so the code avoids hidden global random state. This changes those function signatures and their call sites.
- No Numba (`@jit`, `@njit`, `parallel=True`, `prange`) code is present in this lecture, so no special handling was required.

### DIFF
--- a/lectures/ar1_turningpts.md
+++ b/lectures/ar1_turningpts.md
@@ -123,11 +123,11 @@ we'll plot $.9$ and $.95$ coverage intervals using conditional distribution
 We'll also plot a bunch of samples of sequences of future values and watch where they fall relative to the coverage interval.  
 
 ```{code-cell} ipython3
-def AR1_simulate(rho, sigma, y0, T):
+def AR1_simulate(rho, sigma, y0, T, rng):
 
     # Allocate space and draw epsilons
     y = np.empty(T)
-    eps = np.random.normal(0, sigma, T)
+    eps = rng.normal(0, sigma, T)
 
     # Initial condition and step forward
     y[0] = y0
@@ -137,7 +137,7 @@ def AR1_simulate(rho, sigma, y0, T):
     return y
 
 
-def plot_initial_path(initial_path):
+def plot_initial_path(initial_path, rng):
     """
     Plot the initial path and the preceding predictive densities
     """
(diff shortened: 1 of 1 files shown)


jstac:
@Chihiro2000GitHub 

Sorry for the extra task but would you mind working through this again, looking at the new version of ar1_turningpts?

Please adjust the PR accordingly and make sure there are no conflicts.

Chihiro2000GitHub:
Thanks @jstac, happy to redo this against the new version — I can see the branch conflicts with `main` now.

This is the first time I've had to deal with a conflict, so rather than working it out on my own I wanted to check with you first.

I was planning to start again from the current `main` and redo the migration on top of that, which would mean force-pushing this branch. As far as I know I'm the only one working on this branch, so I don't think it would disturb anyone, but I wanted to make sure. 

Is that okay here, or would you rather I merged `main` into the branch and resolved the conflicts instead?

jstac:
@Chihiro2000GitHub Thanks for your thoughtful response, and for carefully checking.  Yes, the approach is good and you are free to force push.

John.

Chihiro2000GitHub:
@mmcky @HumphreyYang (cc @jstac) — an update on this one.

I reset this branch onto current `main` to redo the migration, and it turns out there is nothing left to migrate: `ar1_turningpts.md` no longer uses `np.random` anywhere. The lecture was moved over to JAX in #584, which landed shortly after I opened this PR.

So I think the right outcome is to close this PR and record `ar1_turningpts.md` in QuantEcon/meta#299 as a JAX lecture with no NumPy random usage to migrate.

How does that sound? Happy to close it myself if you agree.

jstac:
@Chihiro2000GitHub Thanks, true.  I forgot that this is now in JAX.  I'll close the PR.

PR #679, opened 2025-11-07
Refactor optimal growth lectures into unified cake eating series

jstac:
## Summary

This PR restructures five lectures into a coherent **cake eating series** with consistent notation, modern Python patterns, and improved pedagogical flow. The series now progresses logically from deterministic to stochastic problems, introducing increasingly sophisticated solution methods, culminating in a JAX-accelerated implementation.

## Lecture Series Structure

1. **Cake Eating I** (`cake_eating.md`) - Introduction to Optimal Saving
2. **Cake Eating II** (`cake_eating_numerical.md`) - Numerical Methods  
3. **Cake Eating III** (`cake_eating_stochastic.md`) - Stochastic Dynamics ✨ *New framing*
4. **Cake Eating IV** (`cake_eating_time_iter.md`) - Time Iteration ✨ *Renamed*
5. **Cake Eating V** (`cake_eating_egm.md`) - The Endogenous Grid Method ✨ *Renamed*
6. **Cake Eating VI** (`cake_eating_egm_jax.md`) - EGM with JAX ✨ *New*

### 🚀 New JAX Implementation (Cake Eating VI)

Added a new lecture demonstrating high-performance implementation using JAX:

**Key JAX features demonstrated:**
- **Vectorization with `vmap`**: Replaces for loops with efficient vectorized operations
- **JIT compilation**: `@jax.jit` decorator for compiled performance
- **Functional control flow**: `jax.lax.while_loop` instead of Python loops
- **Hardware acceleration**: Supports GPU/TPU through JAX
- **Clean architecture**: Global functions with explicit parameters for JAX compatibility

**Implementation highlights:**
```python
@jax.jit
def solve_model_time_iter(model, σ_init, tol=1e-5, max_iter=1000):
    def condition(loop_state):
        i, σ, error = loop_state
        return (error > tol) & (i < max_iter)
    
    def body(loop_state):
        i, σ, error = loop_state
        σ_new = K(σ, model)
        error = jnp.max(jnp.abs(σ_new - σ))
        return i + 1, σ_new, error
    
    i, σ, error = jax.lax.while_loop(condition, body, initial_state)
    return σ
```

The lecture focuses on JAX implementation patterns while referring readers to Cake Eating V for algorithmic details.

### 🔤 Consistent Notation Throughout

Changed state variable from `y` → `x` in all lectures (III, IV, V) to match the notation in Cake Eating I and II. This includes:
[...]

PR #962, opened 2026-07-12
[optgrowth_fast/ogm.py] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `_static/lecture_specific/optgrowth_fast/ogm.py` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299). The legacy global random-state calls are replaced with an explicit `np.random.default_rng()` generator.

## Details

- The random draw sits inside the `@jitclass` `OptimalGrowthModel.__init__`. Following the QuantEcon jitclass convention, `rng` is passed in as an argument (not stored as a class attribute), and `np.random.seed(seed)` / `np.random.randn(shock_size)` are replaced with `rng.standard_normal(shock_size)`.
- The `seed` parameter is dropped from the constructor; reproducibility now comes from the caller creating `rng = np.random.default_rng(seed)` and passing it in. This changes the `__init__` signature (`rng` added, `seed` removed), so call sites must construct with `OptimalGrowthModel(rng)`.
- Within this repository, no lecture appears to `:load:`/reference `optgrowth_fast/ogm.py`, so there are no in-repo call sites to update here. Any external call sites (e.g. in other lecture repos) will need to construct `rng` and pass it in.
- Verified locally that the sized `rng.standard_normal(shock_size)` call and passing a `Generator` into the `@jitclass` compile, run, and remain reproducible under Numba 0.62.1.

### DIFF
--- a/lectures/_static/lecture_specific/optgrowth_fast/ogm.py
+++ b/lectures/_static/lecture_specific/optgrowth_fast/ogm.py
@@ -14,23 +14,22 @@ opt_growth_data = [
 class OptimalGrowthModel:
 
     def __init__(self,
+                rng,
                 α=0.4,
                 β=0.96,
                 μ=0,
                 s=0.1,
                 grid_max=4,
                 grid_size=120,
-                shock_size=250,
-                seed=1234):
+                shock_size=250):
 
         self.α, self.β, self.μ, self.s = α, β, μ, s
 
         # Set up grid
         self.grid = np.linspace(1e-5, grid_max, grid_size)
 
-        # Store shocks (with a seed, so results are reproducible)
-        np.random.seed(seed)
-        self.shocks = np.exp(μ + s * np.random.randn(shock_size))
+        # Store shocks (drawn from the passed-in rng, so results are reproducible)
+        self.shocks = np.exp(μ + s * rng.standard_normal(shock_size))
 
 
     def f(self, k):


lectures/_static/lecture_specific/optgrowth_fast/ogm.py
```diff
@@ -14,23 +14,22 @@
 class OptimalGrowthModel:
 
     def __init__(self,
+                rng,
```
kp992, line 17: This is the changing the API of how the object is created. Since this code snippet might be directly used in the lecture, it can break that lecture. Maybe it would be good to keep the params of `__init__` method as it and update the function code instead.

mmcky:
Following up on the review discussion (cc @kp992): it turns out no lecture consumes this file any more, so the constructor-signature concern is moot — but for the same reason the file itself is dead code.

Brief history:

- `optgrowth_fast.md` ("Optimal Growth II: Accelerating the Code with Numba") used to `:load:` this file and instantiate `OptimalGrowthModel()` with no arguments — so the concern about changing the `__init__` signature was well-founded for that original consumer.
- #679 (merged Nov 2025) refactored the optimal growth series into the cake eating / `os_*` series with inline `NamedTuple` models and deleted `optgrowth_fast.md`, but left `_static/lecture_specific/optgrowth_fast/` behind.
- Today nothing in this repo references `ogm.py`: no `:load:` directives, and `OptimalGrowthModel` appears nowhere outside this folder.

Given that, the plan is to tidy up the leftover folder rather than migrate its RNG usage — see #1033. Thanks @Chihiro2000GitHub for the careful work here (the migration itself correctly follows the jitclass convention); once the tidy-up lands this PR can be closed, or repurposed into the deletion if you'd like to take that on.

Chihiro2000GitHub:
Thanks @mmcky for looking into the history, and thanks @kp992 for the review. 

@mmcky -- I will work on the clean-up in https://github.com/QuantEcon/lecture-python.myst/issues/1033.

PR #967, opened 2026-07-16
Link ar1_turningpts to ar1_bayes and improve exposition

jstac:
## What

Better links `ar1_turningpts` (Forecasting an AR(1) Process) with `ar1_bayes` (Posterior Distributions for AR(1) Parameters), reduces duplicated exposition, and tightens the lecture's beginning and (previously abrupt) ending.

## Changes

**Cross-linking**
- `ar1_turningpts` now references `ar1_bayes` in the overview, model setup, and posterior section, and treats the posterior construction more briefly rather than re-deriving it.
- Frames the posterior computation as the **conditioning assumption** of `ar1_bayes` — apt here, since the initial path starts from an atypical `y_0 = 10`.
- Added a forward `{seealso}` from `ar1_bayes` back to `ar1_turningpts`.

**Exposition**
- Matched the AR(1) model introduction to the bulleted style used in `ar1_bayes`.
- Added a **Conclusion** and interpretive text after the original/extended Wecker figures, tying results back to the two sources of uncertainty promised in the overview (fixes the abrupt ending).
- Defined the negative turning point statistic `N_t`, which was referenced but never defined.

**Style / cleanup**
- Converted the standalone initial-path figure to a MyST `mystnb` caption (dropped the embedded `ax.set_title`); kept per-subplot titles on the multi-panel grids.
- Fixed typos: `i.i.d.` → `IID`, "we apply we apply", "intergration".

### DIFF
--- a/lectures/ar1_bayes.md
+++ b/lectures/ar1_bayes.md
@@ -584,3 +584,7 @@ A simple rule of thumb:
 
 - use the conditioning assumption when early observations look transient or the starting point may be atypical;
 - use the stationary assumption when you are confident the process has been running near its long-run behaviour.
+
+```{seealso}
+{doc}`ar1_turningpts` builds on the posterior computed here to forecast nonlinear sample-path statistics of an AR(1) process, such as the time until the next turning point.
+```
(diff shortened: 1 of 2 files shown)


jstac:
Superseded by #969, which redoes these edits on top of the numpyro/JAX rewrite of the lecture (#584). This branch was based on the old PyMC version and had become unmergeable.

PR #969, opened 2026-07-16
Link ar1_turningpts to ar1_bayes and improve exposition

jstac:
Better links `ar1_turningpts` (Forecasting an AR(1) Process) with `ar1_bayes` (Posterior Distributions for AR(1) Parameters), and tightens the lecture's beginning and (previously abrupt) ending.

Redone on top of the numpyro/JAX rewrite of `ar1_turningpts` (#584) — supersedes #967, which was based on the old PyMC version and had become unmergeable.

## Changes

**Cross-linking**
- `ar1_turningpts` now references `ar1_bayes` in the overview, model setup, and posterior section, and treats the posterior construction more briefly rather than re-deriving it.
- Frames the posterior computation as the **conditioning assumption** of `ar1_bayes` — apt here, since the initial path starts from an atypical `y_0 = 10`.
- Added a forward `{seealso}` from `ar1_bayes` back to `ar1_turningpts`.

**Exposition**
- Matched the AR(1) model introduction to the bulleted style used in `ar1_bayes` (and `IID` rather than `i.i.d.`).
- Added a **Conclusion** and interpretive text after the original/extended Wecker figures, tying results back to the two sources of uncertainty promised in the overview (fixes the abrupt ending).

## Note

The rewrite (#584) already defines `N_t` and already captions every figure via `mystnb`, so two fixes from #967 are no longer needed here.

### DIFF
--- a/lectures/ar1_bayes.md
+++ b/lectures/ar1_bayes.md
@@ -584,3 +584,7 @@ A simple rule of thumb:
 
 - use the conditioning assumption when early observations look transient or the starting point may be atypical;
 - use the stationary assumption when you are confident the process has been running near its long-run behaviour.
+
+```{seealso}
+{doc}`ar1_turningpts` builds on the posterior computed here to forecast nonlinear sample-path statistics of an AR(1) process, such as the time until the next turning point.
+```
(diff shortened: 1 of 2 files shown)


jstac:
@mmcky This should be good to merge when green.

mmcky:
thanks @jstac -- merging now.

PR #972, opened 2026-07-16
Use vendored data in Hansen-Singleton lectures

nyxst4ck:
## Summary
- read both Hansen-Singleton datasets from their vendored `_static/lecture_specific` paths
- update the surrounding prose to describe the checked-out data source
- remove runtime dependence on raw GitHub URLs

Closes QuantEcon/actions#102

## Validation
- pandas smoke checks: 1982 data `(239, 2)` and 1983 data `(239, 5)`, with expected columns and DatetimeIndex
- executed and built `hansen_singleton_1982` successfully (all cells, HTML)
- executed and built `hansen_singleton_1983` successfully (all cells, HTML)
- git diff --check passes

### DIFF
--- a/lectures/hansen_singleton_1982.md
+++ b/lectures/hansen_singleton_1982.md
@@ -985,19 +985,18 @@ Because the Ken French return is not identical to the original CRSP NYSE value-w
 
 Both this lecture and the companion lecture {doc}`hansen_singleton_1983` use the same data construction.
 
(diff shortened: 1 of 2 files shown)


mmcky:
Thanks so much for opening this PR @nyxst4ck — it's a well-put-together change and the validation notes are much appreciated! 🙏

Unfortunately we can't take this approach, for one main reason: these lectures are also published as standalone Jupyter notebooks (via the Colab launch buttons, backed by [lecture-python.notebooks](https://github.com/QuantEcon/lecture-python.notebooks)), and that environment is a flat directory of `.ipynb` files with no `_static/` folder. A relative local path would raise `FileNotFoundError` for every Colab user of these two lectures, which is why the repo convention (also in `mle`, `ols`, `pandas_panel`) is to read vendored data over the raw GitHub URL — it keeps the notebooks portable.

The HTTP 429 flake in QuantEcon/actions#102 is real, but we think the right fix is on the CI/container side (e.g. retries or caching in the container-validation workflow) rather than changing how the lectures fetch data. We're also moving towards a shared `data-lectures` repository for datasets, which will give us a cleaner long-term answer here.

So we'll close this one — but thanks again for the contribution and for engaging with the QuantEcon lectures. We'd be delighted to see more PRs from you! 😄

PR #976, opened 2026-07-18
[likelihood_ratio_process.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `likelihood_ratio_process.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

A single generator `rng = np.random.default_rng()` is defined once in the setup cell and reused throughout, passed explicitly to the simulation functions.

## Details

- Defined one `rng = np.random.default_rng()` in the imports/setup cell and reused it across all main-text cells and simulation functions.
- Migrated the legacy calls: `np.random.beta` → `rng.beta`, `np.random.rand(N)` → `rng.random(N)`, `np.random.rand(N, T)` → `rng.random((N, T))`, and `np.random.choice(...)` → `rng.choice(...)`.
- Threaded `rng` as an explicit argument through the simulation functions and their wrappers (`simulate_likelihood_ratios`, `simulate_sequences`, `protocol_1`/`protocol_2` and their `compute_*`/`analyze_*` wrappers, `simulate_markov_chain`, `analyze_markov_chains`, `compute_markov_selection_error`) and updated the call sites.
- No fixed seed was introduced; the original code did not seed, so simulation behaviour is unchanged.

## Note for reviewers

- `simulate_likelihood_ratios` is decorated with `@jit` but has no `parallel=True`/`prange`, so `rng` is passed in as an explicit argument (`rng.beta`) rather than lifting the draws into a precomputed array. Verified locally that this compiles and runs under Numba 0.62.1.
- `rng` is defined once in the setup cell rather than at the first point of use, because it is shared across many far-apart sections; a single top-level generator seemed clearer and easier to locate than repeating `default_rng()` per block.
- `mc.simulate` inside `simulate_markov_chain` uses QuantEcon's own internal RNG and is not an `np.random.*` call, so it is left unchanged (out of scope for this migration).

### DIFF
--- a/lectures/likelihood_ratio_process.md
+++ b/lectures/likelihood_ratio_process.md
@@ -63,6 +63,8 @@ from scipy.stats import beta as beta_dist
 import pandas as pd
 from IPython.display import display, Math
 import quantecon as qe
+
+rng = np.random.default_rng()
 ```
 
 ## Likelihood Ratio Process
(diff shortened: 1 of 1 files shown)


kp992 (approved):

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/228
**Files synced** (1):
- `lectures/likelihood_ratio_process.md`

PR #977, opened 2026-07-18
[likelihood_bayes.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `likelihood_bayes.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

The lecture used the `@jit`-based `set_seed()` / `np.random.seed` idiom for reproducibility. It is replaced with an explicit seeded generator `rng = np.random.default_rng(142857)` that is passed into the simulation functions.

## Details

- Replaced the `@jit def set_seed(): np.random.seed(142857)` mechanism with `rng = np.random.default_rng(142857)`. The lecture reset the seed at two points (setup cell and before the mixture path), so `rng` is (re)created at both points to preserve that behaviour.
- Migrated `np.random.beta` → `rng.beta` and `np.random.rand()` → `rng.random()`, and threaded `rng` as an explicit argument through `simulate`, `simulate_mixture_path`, `martingale_simulate` (and its `fraction_0_1` / `create_table` wrappers), and `compute_cond_var`, updating the call sites.
- Kept the existing fixed seed (`142857`); the original lecture already seeded for reproducibility, so no new seed was introduced.

## Note for reviewers

- **`compute_cond_var` (the `prange` case):** the original used `@jit` with a `prange` loop containing the RNG draws, but there is no `parallel=True` anywhere in the file, so the loop already ran serially. Rather than lifting the draws out (which, since the loop's only content was the sampling, would dissolve the loop entirely and defeat the "pass a pre-drawn array into a jitted `prange` kernel" pattern), I changed `prange` → `range` so the function becomes an ordinary serial `@jit` function like the others, and passed `rng` in as an argument. This changes no runtime behaviour (it was already serial). If you intended this to be parallelized, the RNG should instead be drawn outside and passed in — happy to switch to that. `prange` is now unused, so I also removed it from the `from numba import ...` line.
- The other jitted functions (`simulate`, `simulate_mixture_path`, `martingale_simulate`) have no `parallel=True`/`prange`, so `rng` is passed in explicitly (Case 2). Verified locally that passing a `Generator` into these `@jit` functions and calling `rng.beta` / `rng.random` compiles and runs under Numba 0.62.1.
- The fixed seed `142857` is retained, so results remain reproducible.


lectures/likelihood_bayes.md
```diff
@@ -512,8 +509,8 @@ prior_params = [(1, 3), (1, 1), (3, 1)]
 prior_means = [a/(a+b) for a, b in prior_params]
 
 # Generate one path of observations from the mixture
-set_seed()
-w_mix = simulate_mixture_path(x_true, T_mix)
+rng = np.random.default_rng(142857)
```
kp992, line 512:
```suggestion
rng = np.random.default_rng(seed=142857)
```

It helps readers to understand what this random number is

lectures/likelihood_bayes.md
```diff
@@ -989,15 +986,15 @@ Then we'll plot it.
 
 ```{code-cell} ipython3
 @jit
-def compute_cond_var(pi, mc_size=int(1e6)):
+def compute_cond_var(pi, rng, mc_size=int(1e6)):
     # create monte carlo draws
     mc_draws = np.zeros(mc_size)
 
-    for i in prange(mc_size):
```
kp992, line 996: prange is removed by mistake?

kp992 (changes requested):

mmcky:

lectures/likelihood_bayes.md
mmcky: Thanks @kp992 — I checked the convention across the series before deciding on this one: on `main`, `default_rng(...)` is called with a positional seed 135 times versus once with `seed=`, so the positional form is the established style in these lectures. I'd prefer to keep it positional here for consistency (a series-wide move to `seed=` would be better proposed as its own style change).

Also worth noting: there are two `default_rng(142857)` call sites in this lecture (the setup cell and this reset before the mixture path), so if we did adopt `seed=`, both would need it — applying the suggestion only here would leave them inconsistent.

lectures/likelihood_bayes.md
mmcky: Not a mistake — it's deliberate (see the "Note for reviewers" in the PR description), and the history backs it up: `parallel=True` has never appeared anywhere in this file (`git log -S 'parallel=True' -- lectures/likelihood_bayes.md` returns nothing). `prange` was introduced already-serial in 86f176b (Dec 2022) — #284 only added the then-missing import — and Numba treats `prange` as ordinary `range` unless the decorator sets `parallel=True`. So this loop has always executed serially, and `prange` → `range` is behavior-neutral; it just makes the code say what it actually does. (Empirically the full 40-point grid runs in ~12s serial, so there's no practical performance concern.)

Keeping `prange` wouldn't be viable under the new API: a single `Generator` shared across threads in a parallel region isn't thread-safe. If we ever want this loop genuinely parallel, the right pattern is to pre-draw the samples outside the jitted kernel (or use per-thread generators via `rng.spawn()`) — best done as a separate change. Since `prange` is now unused in the file, removing it from the import line is correct too.

Do you agree?

mmcky (approved):

Chihiro2000GitHub:
Thanks @mmcky — sorry these were mine to answer and you got there first. 

Both replies are exactly what I would have said, so I have nothing to add; much appreciated, including the conflict resolution. 

I'll take it from here and check the rebuilt preview once the build finishes.

mmcky:
> Thanks @mmcky — sorry these were mine to answer and you got there first.
> 
> Both replies are exactly what I would have said, so I have nothing to add; much appreciated, including the conflict resolution.
> 
> I'll take it from here and check the rebuilt preview once the build finishes.

Thanks so much @Chihiro2000GitHub. Appreciate the review.

Chihiro2000GitHub:
@mmcky -- I've now gone through the rebuilt preview myself and it looks good.

The one visible difference from `main` is Fig. 29.3 (`fig-posterior-mixture`), where $\pi_t$ converges to 1 noticeably earlier. The migration changes the draw sequence, so the realized path behind that figure is different. It doesn't contradict the prose that follows ("Evidently, $\pi_t$ converges to 1"), so I'd leave it as is.

Everything else builds and reads exactly as it should, so no concerns at all from my side.

mmcky:
thanks @Chihiro2000GitHub. 

@kp992 if you disagree -- feel free to open an issue. I'll merge this now.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/248
**Files synced** (1):
- `lectures/likelihood_bayes.md`

kp992:
Thanks @mmcky and @Chihiro2000GitHub. I am fine with the current PR as it is aligning with the style guide.

PR #978, opened 2026-07-18
[mix_model.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `mix_model.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

The lecture used the module-level `np.random.seed` / `@jit set_seed()` idiom for reproducibility. It is replaced with an explicit seeded generator `rng = np.random.default_rng(142857)` that is passed into the simulation functions.

## Details

- Replaced the `np.random.seed(142857)` / `@jit def set_seed()` mechanism with `rng = np.random.default_rng(142857)`. The lecture reset the seed at two points (setup cell and before the long-horizon paths), so `rng` is (re)created at both points to preserve that behaviour.
- Migrated `np.random.beta` → `rng.beta` and `np.random.rand` → `rng.random`, and threaded `rng` as an explicit argument through `simulate`, `draw_lottery`, `draw_lottery_MC`, `simulate_mixed`, and `plot_π_seq`, updating all call sites.
- Kept the existing fixed seed (`142857`); the original lecture already seeded for reproducibility, so no new seed was introduced.

## Note for reviewers

- **`π_lim` (used with `np.vectorize`):** `π_lim` is wrapped by `π_lim_v = np.vectorize(π_lim)` and calls `simulate_mixed`. Adding `rng` as a positional argument would make `np.vectorize` try to broadcast it, so I left `π_lim`'s signature unchanged and let it reference the shared module-level `rng` inside. This reuses the same single generator, just via closure rather than an argument. The explicit-argument alternative would be:
  ```python
  def π_lim(α, rng, T=5000, π_0=0.4):
      ...
      l_arr = simulate_mixed(α, rng, T, N=1)[0]
      ...
  π_lim_v = np.vectorize(π_lim, excluded=['rng'])
  # called as π_lim_v(α_arr_x, rng=rng)
  ```
  Happy to switch to that if you prefer keeping `rng` explicit everywhere.
- **Prose at "method 1":** the pseudo-code bullet previously said "use the `numpy.random.choice` function to flip an unfair coin". I updated it to "use the `rng.choice` method …" to match the Generator API. Note the actual `draw_lottery` code implements the coin flip with `rng.random() <= p` rather than `rng.choice`, so this is a conceptual description — let me know if you'd rather word it differently.
- **`x=0.5` convergence figure:** under the new generator the realized 200-step path in the "three different prior beliefs" figure (`T_mix = 200`) converges less visibly to `x=0.5` than before, though the text says the posterior means "eventually converge". The theoretical statement still holds; if you'd like the figure to match more tightly, `T_mix` could be increased. I left it unchanged to keep this PR focused on the RNG migration.
- The `@jit` functions (`simulate`, `draw_lottery`) have no `parallel=True`/`prange`, so `rng` is passed in explicitly (Case 2). The numpyro/JAX MCMC uses `jax.random.PRNGKey`, which is not an `np.random.*` call and is left unchanged. Verified locally that the notebook compiles and runs under Numba 0.62.1.

### DIFF
--- a/lectures/mix_model.md
+++ b/lectures/mix_model.md
@@ -141,12 +141,7 @@ from numpyro.infer import MCMC, NUTS
 import jax.numpy as jnp
 from jax import random
 
-np.random.seed(142857)
-
-@jit
-def set_seed():
-    np.random.seed(142857)
-set_seed()
+rng = np.random.default_rng(142857)
 ```
 
 Let's use Python to generate two beta distributions
@@ -172,7 +167,7 @@ g = jit(lambda x: p(x, G_a, G_b))
 :hide-output: false
 
 @jit
-def simulate(a, b, T=50, N=500):
+def simulate(a, b, rng, T=50, N=500):
     '''
     Generate N sets of T observations of the likelihood ratio,
     return as N x T matrix.
@@ -184,7 +179,7 @@ def simulate(a, b, T=50, N=500):
     for i in range(N):
 
         for j in range(T):
-            w = np.random.beta(a, b)
+            w = rng.beta(a, b)
             l_arr[i, j] = f(w) / g(w)
 
     return l_arr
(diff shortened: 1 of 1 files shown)


jstac:
Many thanks @Chihiro2000GitHub , nice work.  There is, however, an issue of overlapping PRs.  I'll write separately to explain.

Issue #882, opened 2026-05-28
rs_inventory_q / inventory_q: misleading 'Converged' print + snapshot off-by-one in Q-learning loop


mmcky:
While reviewing [`lectures/rs_inventory_q.md`](https://github.com/QuantEcon/lecture-python.myst/blob/main/lectures/rs_inventory_q.md) as part of mirroring it into [`QuantEcon/lecture-dp`](https://github.com/QuantEcon/lecture-dp/pull/20), Copilot flagged three things — all real, none blocking, but worth a small follow-up here.

## Issue 1 — \`solve_rs_inventory_model\` prints "Converged" unconditionally

\`rs_inventory_q.md\` (and the same pattern in \`inventory_q.md\` line 331):

```python
def solve_rs_inventory_model(v_init, model, max_iter=10_000, tol=1e-6):
    v = v_init.copy()
    i, error = 0, tol + 1
    while i < max_iter and error > tol:
        new_v = T_rs(v, model)
        error = np.max(np.abs(new_v - v))
        i += 1
        v = new_v
    print(f"Converged in {i} iterations with error {error:.2e}")
    ...
```

If the loop exits because \`i == max_iter\` (before \`error <= tol\`), the print still claims convergence. With the default 10k cap that won't bite at the configured parameters, but it's misleading for anyone who tweaks settings.

Suggestion: only print "Converged" when \`error <= tol\`, otherwise warn (or raise). Same fix applies in \`inventory_q.md\`.

## Issue 2 — snapshot recorded at the *start* of the loop iteration

\`rs_inventory_q.md\` (\`q_learning_rs_kernel\`):

```python
for t in range(n_steps):
    # Record policy snapshot if needed
    if snap_idx < n_snaps and t == snapshot_steps[snap_idx]:
        snapshots[snap_idx] = greedy_policy_from_q_rs(q, K)
        snap_idx += 1
    ...
    q[x, a] = (1 - α) * q[x, a] + α * target
    ...
```

The snapshot at index \`t\` captures \`q\` *before* the update for step \`t\`. So a "snapshot at step n" actually reflects \`q\` after \`n\` completed updates only if \`t = n\` is visited *after* the n-th update — which requires running \`n + 1\` iterations.

That's exactly what the caller does, but it's load-bearing for the narrative…

## Issue 3 — narrative vs. code mismatch in "Running Q-learning"

\`rs_inventory_q.md\` line 658:

[...]

PR #1013, opened 2026-07-28
[rs_inventory_q.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `rs_inventory_q.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

All of the random draws happen inside `@numba.jit(nopython=True)` functions, so the two of them now take a `rng` argument and the generator is built by the calling Python code.

## Related PRs and issues

I checked for open PRs and issues related to this lecture. No open PR modifies `rs_inventory_q.md`. Issue #882 concerns the convergence message and the snapshot timing in the Q-learning loop, and this PR does not attempt to address it.

## Details

- `sim_inventories` takes `rng` in place of `seed`, and draws with `rng.geometric(p)`.
- `q_learning_rs_kernel` takes `rng` in place of `seed`, and draws with `rng.geometric(p)`, `rng.integers(...)` and `rng.random()`.
- The `q_learning_rs` wrapper is ordinary Python, so it keeps its `seed=1234` argument and builds the generator itself. Its call site is unchanged.
- The existing seeds (`0`, `1234`, `5678`) are all kept, and no new seed was introduced.
- Both simulation figures build a fresh `np.random.default_rng(sim_seed)` for each call, so the panels still face the same demand sequence, as they did when each call re-seeded. This matters for the figure comparing risk sensitivities and for the figure comparing the Q-learning snapshots against the VFI policy.
- Numba supports `geometric`, `integers` and `random` on a `Generator` passed in as an argument, but does not support constructing one inside a jitted function, which is why the generator is built by the caller. There is no `prange` or `parallel=True` in this lecture, so a single generator is safe here.
- No prose changes were needed.
- A full local build completed successfully and `rs_inventory_q.md` executed without errors.

## Note for reviewers

Both jitted functions take `rng` in place of `seed`, which changes their signatures. `q_learning_rs` keeps its `seed` argument so the call in the lecture is unchanged, but `sim_inventories` is called directly and now receives a generator at each call site.

### DIFF
--- a/lectures/rs_inventory_q.md
+++ b/lectures/rs_inventory_q.md
@@ -332,13 +332,12 @@ We simulate inventory dynamics under the optimal policy for the baseline $\gamma
 
 ```{code-cell} ipython3
 @numba.jit(nopython=True)
-def sim_inventories(ts_length, σ, p, X_init=0, seed=0):
+def sim_inventories(ts_length, σ, p, rng, X_init=0):
     """Simulate inventory dynamics under policy σ."""
-    np.random.seed(seed)
     X = np.zeros(ts_length, dtype=np.int32)
     X[0] = X_init
     for t in range(ts_length - 1):
-        d = np.random.geometric(p) - 1
+        d = rng.geometric(p) - 1
         X[t+1] = max(X[t] - d, 0) + σ[X[t]]
     return X
 ```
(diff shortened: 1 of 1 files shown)


kp992 (approved):

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/215
**Files synced** (1):
- `lectures/rs_inventory_q.md`

PR #1040, opened 2026-08-24
[lq_robust_bewley] verify observational equivalence

mmcky:
Verify observational equivalence, fix sigma normalisation

The observational-equivalence demonstration was circular: it imposed the benchmark consumption rule on every type and then reported that the resulting paths coincided. The loop body never used the type index, so the reported difference of zero was forced and would have printed identically had the proposition been false.

Replace it with a genuine test. Each type's robust LQ problem is now solved separately at its own (sigma_i, beta_i), facing the common market rate R = 1/beta, using the risk-sensitive solver of robust_permanent_income. Every coefficient of every type's rule matches the benchmark to 1e-12 or better. A falsification step moves each discount factor one percent off the locus and recovers deviations of order 0.1 to 1, so the agreement is not an artifact. The path simulation now runs each type under its own solved rule, and the cross-section experiment gives each agent its own solved problem rather than drawing a type and discarding it.

Fix a normalization inconsistency this uncovered. The stated objective carried a factor of 1/2 while the sigma in the observational-equivalence locus is normalized to a period return of -(c-b)^2; numerically the two differ by exactly a factor of two in the implied locus.

Also:

- give the breakdown point an interpretation: zeta(sigma_lo) = 1/beta = R, so it is where the feared growth in marginal utility reaches the gross interest rate and the worst-case objective ceases to converge
- correct zeta_i > 1 to zeta_i >= 1, with equality at sigma_i = 0
- state Proposition part 2 as existence rather than uniqueness, since the locus is itself constructed at R = 1/beta
- explain why heterogeneous discounting does not degenerate the wealth distribution here
- raise the detection-error target in the last exercise from 0.20 to 0.25; the DEP at the breakdown point is 0.1999, so the old target made the exercise's conclusion depend on the random seed
- align the state timing label with eq:rbew-law

### DIFF
--- a/lectures/lq_robust_bewley.md
+++ b/lectures/lq_robust_bewley.md
@@ -96,7 +96,9 @@ c_t
 (w_{t+1} + v_{t+1})
 $$ (eq:rbew-law)
 
-The objective is $\mathbb{E}_0 \sum_{t=0}^\infty \beta^t\bigl[-(c_t - b)^2/2\bigr]$, which is the HST criterion with $\sigma = 0$ and a constant bliss level $b_t \equiv b$.
+The objective is $\mathbb{E}_0 \sum_{t=0}^\infty \beta^t\bigl[-(c_t - b)^2\bigr]$, which is the HST criterion with $\sigma = 0$ and a constant bliss level $b_t \equiv b$.
+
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @thomassargent30

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/269
**Files synced** (1):
- `lectures/lq_robust_bewley.md`

PR #1044, opened 2026-08-26
[qr_decomp] Some updates to code and wording

### LINKED ISSUE #1043: [qr_decomp] Some suggestion on the code an wording
Hi @jstac 
I was reading through the `qr_decomp` lecture with codex and have some suggestions for the code and wording. 
I've reviewed each of the suggestions myself.

I'd be grateful if you could take a look when you have a chance.

### Matrix Factorization

- [x] At lines 31–45, define the dimensions of a reduced QR decomposition and distinguish between an upper-triangular and an upper-trapezoidal matrix. For example, if `A` is an `n × m` matrix and `k = min(n, m)`, then

$$
A = QR,\qquad
Q \in \mathbb{R}^{n \times k},\qquad
R \in \mathbb{R}^{k \times m},\qquad
Q^\top Q = I_k.
$$

  Here, `R` is upper triangular when it is square and upper trapezoidal when it is rectangular, meaning that all entries below its main diagonal are zero.

- [x] At line 42, use `\top` rather than `T` for the transpose.

- [x] Change the section heading to sentence case: “Matrix factorization.”

- [x] Add terminal punctuation to the sentences at lines 47 and 49.

### Gram-Schmidt process

- [x] At line 55, qualify the uniqueness claim. A nonsingular square matrix does not have a unique QR factorization without a sign convention. A standard convention is to require the diagonal entries of `R` to be positive.

- [x] Apply the style-guide distinction between bold and italic consistently. Bold should identify terms being defined, while italics should indicate ordinary emphasis. For example, `normalize` and `orthogonalize` can remain bold if they are being introduced as defined steps, whereas words such as `square`, `columns`, and the later occurrences of `eigenvalues` should use italics or plain text.

- [x] Remove math formatting from the prose term “QR” at line 55.

#### Gram-Schmidt process for square A

- [x] At lines 128–131, correct the displayed definition of `Q`. It should contain the orthonormal vectors, without also being equated to the original columns of `A`:

$$
Q =
\begin{bmatrix}
e_1 & e_2 & \cdots & e_n
\end{bmatrix}.
$$

- [x] Replace the Unicode middle dot used in mathematical expressions with `\cdot`.

- [x] Replace the `array` and `matrix` environments used to display matrices with `bmatrix`, following the QuantEcon matrix-formatting convention.

- [x] At line 120, correct “decomposision” to “decomposition.”

#### A not square

- [x] Describe the resulting `n × m` matrix `R` as upper trapezoidal rather than upper triangular.

- [x] Remove math formatting from the prose term “QR” at line 144.

### Some Code

- [x] At line 173, initialize the random number generator with a fixed seed so that all later examples are reproducible, for example:

```python
rng = np.random.default_rng(1234)
```

- [x] Rename `QR_Decomposition` to `qr_decomposition` and update its uses.

- [x] The `QR_Decomposition` function at lines 177–199 supports square and wide matrices, where $m\ge n$, but not tall matrices, where $n>m$ . We could add a clarification sentence immediately before the function stating its supported dimensions and its assumption that the first n columns are linearly independent.

- [x] At line 192, correct “vetor” to “vector.”

- [x] At lines 206–235, revise the sign-normalization convention. Requiring a positive diagonal in `Q` is nonstandard, and `diag_sign(Q)` can destroy the factorization because `np.sign(0)` returns zero. For example:

```python
A = np.array([
    [0.0, 1.0],
    [1.0, 0.0]
])
```

  If a diagonal entry of `Q` is zero, the resulting diagonal sign matrix is singular and multiplication by it zeros a column of `Q` and a row of `R`. The usual convention is to choose signs from the diagonal of `R`, while mapping any zero sign to `1`, so that the diagonal of `R` is nonnegative and the identity

$$
A = QR
$$

  is preserved.

- [x] At line 206, remove the duplicated word in “There can be be sign differences.”

- [x] Change the section heading to sentence case: “Some code.”

### Example

- [x] At lines 262–281, capitalize the package name consistently as “SciPy” when printed out or used in the text including the printed labels.

- [x] The stopping and return logic in `QR_eigvals ` at lines 337–350 can return incorrect eigenvalues. The routine treats a small change between consecutive iterates as convergence, but a stationary iterate need not be diagonal. For example, the matrix [[0, 1], [1, 0]] can remain unchanged, causing diff == 0, after which the function returns [0, 0] instead of [-1, 1]. The function also returns the diagonal unconditionally when maxiter is reached.

- [x] At lines 334–343, convert the working matrix to floating-point storage. `np.copy(A)` preserves an integer input dtype, so later floating-point iterates are truncated during in-place assignment. For example:

```python
A = np.array([
    [2, 1],
    [1, 2]
])
```

  A suitable initialization would explicitly construct a floating-point working array.

- [x] Remove the maintainer TODO at lines 322–324 because it is visible in the Colab version. When this section is revised, the numbered procedure could be migrated to a `prf:algorithm` directive.

- [x] Replace the manually bolded remark at line 326 with a `prf:remark` directive. Its statement should also be revised alongside the decision about the algorithm’s intended scope.

- [x] Rename `QR_eigvals` to `qr_eigvals`.

- [x] At lines 353–369, either change the text to say NumPy The current code calls `np.linalg.eigvals`, not `scipy.linalg.eigvals`.

- [x] Use a deterministic and deliberately supported example rather than an arbitrary random general matrix. Seeding the generator makes the example reproducible, but it does not ensure that the sampled matrix lies in the class the simplified algorithm can handle.

- [x] Change the section heading to sentence case: “Using QR decomposition to compute eigenvalues.”

- [x] Remove math formatting from the prose term “QR” at line 314.

### QR and PCA

- [x] Remove math formatting from “QR” in the heading and at line 374. The heading should be plain text: “QR and PCA.”

- [x] Replace prime notation for matrix transposes with `\top` throughout this section. 

- [x] At lines 378–380, use the standard normal-distribution notation

$$
N(\mu, \Sigma)
$$

  rather than a calligraphic `N`, and replace `n > > k` with

$$
n \gg k.
$$

according to the style guide.

- [x] At line 389, remove the duplicated “that.”

- [x] At line 407, correct the code comment. `rng.multivariate_normal(..., size=n)` returns an `n × k` array in which each row, not each column, is a multivariate-normal draw.

- [x] At lines 451–459, replace the diagonal-based eigenvector sign adjustment. `diag_sign` can zero a column when an eigenvector has a zero diagonal entry, and separate diagonal signs do not reliably align corresponding eigenvectors. A sign-invariant comparison could instead inspect

```python
np.abs(np.diag(P.T @ QP_tilde))
```

What do you think? Happy to put up a PR. 

Best,  
Longye

longye-tian:
Hi @jstac,

This PR updates the `qr_decomp` lecture in line with #1043. In particular, 

- update the matrix dimensions and upper-trapezoidal structure.
- updates typos, headings and format according our style guide. 
- updates the QR sign convention to use the diagonal of `R` so a zero diagonal will not cause problems.
- update the code on integer to floating point in the QR eigenvalue iteration.

Best,  
Longye

### DIFF
--- a/lectures/qr_decomp.md
+++ b/lectures/qr_decomp.md
@@ -43,15 +43,15 @@ The matrix $R$ has zeros below its main diagonal.
 
 Hence, $R$ is upper triangular when $m = n$ and upper trapezoidal when $m > n$.
 
-We'll use a **Gram-Schmidt process** to compute a  QR decomposition 
+We'll use a **Gram-Schmidt process** to compute a  QR decomposition.
 
-Because doing so is so educational, we'll  write our own Python code to do the job
+Because doing so is so educational, we'll  write our own Python code to do the job.
 
 ## Gram-Schmidt process
 
 We'll start with a **square** matrix $A$.
 
-If a square matrix $A$ is nonsingular, then a $QR$ factorization is unique.
+If a square matrix $A$ is nonsingular, then a QR factorization is unique.
 
 We'll deal with a rectangular matrix $A$ later.
 
@@ -140,7 +140,7 @@ $$
 
 Now suppose that $A$ is an $n \times m$ matrix where $m > n$.  
 
-Then a $QR$ decomposition is
+Then a QR decomposition is
 
 $$
 A= \left[ \begin{array}{c|c|c|c} a_1 & a_2 & \cdots & a_m \end{array} \right]=\left[ \begin{array}{c|c|c|c} e_1 & e_2 & \cdots & e_n \end{array} \right]
@@ -310,7 +310,7 @@ Here is the algorithm:
 
 2. Form $A_1 = R_0 Q_0 $ . Note that $A_1$ is similar to $A_0$ (easy to verify) and so has the same eigenvalues.
 
-3. Form $A_1 = Q_1 R_1$ (i.e., form the $QR$ decomposition of $A_1$).
+3. Form $A_1 = Q_1 R_1$ (i.e., form the QR decomposition of $A_1$).
 
 4. Form $ A_2 = R_1 Q_1 $ and then $A_2 = Q_2 R_2$  .
 
@@ -370,7 +370,7 @@ sorted(np.linalg.eigvals(A))
 
 ## QR and PCA
 
-There are interesting connections between the $QR$ decomposition and principal components analysis (PCA).
+There are interesting connections between the QR decomposition and principal components analysis (PCA).
 
 Here are  some.
 
--- a/lectures/qr_decomp.md
+++ b/lectures/qr_decomp.md
@@ -148,7 +148,9 @@ A= \begin{bmatrix} a_1 & a_2 & \cdots & a_m \end{bmatrix}=\begin{bmatrix} e_1 & 
 \\ 0 & 0 & \cdots & a_n\cdot e_n & a_{n+1}\cdot e_n & \cdots & a_{m}\cdot e_n \end{bmatrix}
 $$
 
-which implies that
+Here, $R$ is an $n \times m$ upper trapezoidal matrix.
+
+This implies that
 
 \begin{align*}
 a_1 & = (a_1\cdot e_1) e_1 \cr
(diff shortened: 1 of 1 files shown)


jstac:
Great work @longye-tian , much appreciated!

Everything is nice and I'm going to merge.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python.zh-cn/pull/277
**Files synced** (1):
- `lectures/qr_decomp.md`

PR #1050, opened 2026-09-03
[sir_model] update code and text

### LINKED ISSUE #1049: [sir_model] Some suggestions on code and wording
Hi @jstac 

I was reading through the `sir_model` lecture and noted some possible updates. 

### Overview and model description

- [x] Change both occurrences of “Modeling COVID 19” to “Modeling COVID-19,” including the index entry to make them consistent with the main text and also consistent with Wikipidea.

- [x] update the “COVID-19 Working papers and code” link to `https://sites.google.com/site/andyatkeson/home/covid-work`, currently, this link goes to the personal website of Andrew Atkeson instead of this paper and code. 

- [x] update the four-state model as an SEIR (Susceptible-Exposed-Infected-Removed) model rather than an SIR model. In this lecture, we uses 4 state instead of three states as the name SIR suggested.

- [x] **[qe-writing-006]** Use sentence case for the lower-level headings at `lectures/sir_model.md:68,82,236,284,347`:

  - `The SEIR model`
  - `Time path`
  - `Experiment 1: constant R0 case`
  - `Experiment 2: changing mitigation`
  - `Ending lockdown`

### State definitions and assumptions

- [x] At line 74, we could update the compartment symbols in math delimiters: susceptible ($S$), exposed ($E$), infected ($I$), and removed ($R$). Currently, we have `(R)` for the removed ones  and markdown rendered this as a trademark `®`.

- [x] At line 76-80, replace the “Comments:” label with a `prf:assumption` environment since they are all assumptions used later in this lecture:

````markdown
```{prf:assumption}
- Those in state $R$ were previously infected and have either recovered or died.
- Those who have recovered are assumed to have acquired immunity.
- Those in the exposed group are not yet infectious.
```
````

### Time path and parameters

- [x] At line 86-87, replace the claim that positive transmission and $i(0)>0$ imply everyone is eventually infected as this is not true in this model And we didn't introduce the meaning of $i$ yet. Suggested text:

  > When the effective reproduction number exceeds one, infections initially increase but eventually decline as the susceptible population falls.

- [x] Apply the following small prose corrections at line 91,94,112,114,131:
  - Change “lower case” to “lowercase.”
  - Add the missing period to the $\sigma$ definition.
  - Capitalize “The dot symbol.” as it is a beginning of a new sentence.
  - Write “for a suitable definition of $F$.” by adding the missing article "a"

- [x] **[qe-math-001]** At line 103, 105 , replace the raw Unicode symbols inside the math directive with `\sigma` and `\gamma`.

- [x] At line 113,  call $\gamma$ the **removal rate**, since the transition includes both recovery and death. Currently it is calle d recovery rate which doesn't reflect that this also include dealth. 

### Implementation

- [x] At line 192-199, explain the initial-condition calibration. With a population of 330 million, $i_0=10^{-7}$ represents 33 initially infected people, $e_0=4i_0$ represents 132 exposed people, and $s_0=1-i_0-e_0$ sets the initial removed fraction to zero. This makes the calibrated number economically meaningful. 

- [x] At line 207, format `odeint` as inline code. 

### Experiments and figures

- [x] At line 244, render the zero in the legend variable as a subscript:

  ```python
  labels = [f'$R_0 = {r:.2f}$' for r in R0_vals]
  ```

- [x] **[qe-fig-008]** At line 261,317, add `lw=2` to the line plots.

- [x] At line 256-265,313-320, add axis labels and units so that reader can know the unit of the plot. The shared `plot_paths` helper could accept a y-axis label, while all relevant plots use `days` on the x-axis. Quantity-specific y-axis labels should distinguish:

  - current cases as a fraction of the population;
  - cumulative cases as a fraction of the population;
  - active infections as a fraction of the population;
  - cumulative deaths;
  - deaths per day.

  The standalone mitigation plot should use `$R_0$` on the y-axis.

- [x] At line 301-309, explain that $\eta$ is measured per day because $t$ is measured in days. Its reciprocal is the characteristic adjustment period, so the specified values correspond to adjustment periods of 5, 10, 20, 50, and 100 days.

- [x] After line 345, add some economic interpretation, currently it finishes with the plot. 

  > Faster implementation of mitigation mainly delays the infection peak, with a smaller effect on its height.

### Ending lockdown and mortality

- [x] At line 383, describe the plotted quantity as a fraction rather than a number:

  > Here is the fraction of the population with an active infection:

- [x] After the active-infections plot at line 386, add some economic interpretation. 

  > Both scenarios produce approximately the same infection peak, but the longer lockdown delays it.

- [x] At line 391, add the missing period after “Suppose that 1% of cases result in death.”

- [x] At line 397-401, correct the cumulative-deaths calculation. Since $c(t)=i(t)+r(t)$, the current calculation counts a fraction of active infections as deaths and begins with 250 cumulative deaths even though the initial removed fraction is zero. Cumulative deaths should be $\nu r(t)N=\nu[c(t)-i(t)]N$:

  ```python
  paths = [(c_path - i_path) * ν * pop_size
           for c_path, i_path in zip(c_paths, i_paths)]
  ```

  This is consistent with the daily-deaths formula $\nu\gamma i(t)N$. Label the y-axis `cumulative deaths`.

- [x] At line 404-408, change the introduction to:

  > This is the number of deaths per day:

instead of using daily death rate as the following plot has y-axis as the number of death per day as unit and keep the existing calculation and label the y-axis `deaths per day`.

- [x] At line 411, write “the peak of the curve.”

What do you think? I'd be grateful if you could take a look when you have a chance.
Happy to put up a PR. 

Best,  
Longye

longye-tian:
Hi @jstac 

This PR updates the `sir_model` lecture according to #1049 

Best,
Longye

### DIFF
--- a/lectures/sir_model.md
+++ b/lectures/sir_model.md
@@ -17,7 +17,7 @@ kernelspec:
 </div>
 ```
 
-# {index}`Modeling COVID 19 <single: Modeling COVID 19>`
+# {index}`Modeling COVID-19 <single: Modeling COVID-19>`
 
 ```{contents} Contents
 :depth: 2
@@ -31,12 +31,12 @@ provided by [Andrew Atkeson](https://sites.google.com/site/andyatkeson/).
 See, in particular
 
 * [NBER Working Paper No. 26867](https://www.nber.org/papers/w26867)
-* [COVID-19 Working papers and code](https://sites.google.com/site/andyatkeson/home?authuser=0)
+* [COVID-19 Working papers and code](https://sites.google.com/site/andyatkeson/home/covid-work)
 
 The purpose of his notes is to introduce economists to quantitative modeling
 of infectious disease dynamics.
 
-Dynamics are modeled using a standard SIR (Susceptible-Infected-Removed) model
+Dynamics are modeled using a standard SEIR (Susceptible-Exposed-Infected-Removed) model
 of disease spread.
 
 The model dynamics are represented by a system of ordinary differential
@@ -65,9 +65,9 @@ from scipy.integrate import odeint
 
 This routine calls into compiled code from the FORTRAN library odepack.
 
-## The SIR Model
+## The SEIR model
 
-In the version of the SIR model we will analyze there are four states.
+In the version of the SEIR model we will analyze there are four states.
 
 All individuals in the population are assumed to be in one of these four states.
 
@@ -79,7 +79,7 @@ Comments:
 * Those who have recovered are assumed to have acquired immunity.
 * Those in the exposed group are not yet infectious.
 
-### Time Path
+### Time path
 
 The flow across states follows the path $S \to E \to I \to R$.
 
@@ -233,7 +233,7 @@ grid_size = 1000
 t_vec = np.linspace(0, t_length, grid_size)
 ```
 
-### Experiment 1: Constant R0 Case
+### Experiment 1: constant R0 case
 
 Let's start with the case where `R0` is constant.
 
@@ -281,7 +281,7 @@ Here are cumulative cases, as a fraction of population:
 plot_paths(c_paths, labels)
 ```
 
-### Experiment 2: Changing Mitigation
+### Experiment 2: changing mitigation
 
 Let's look at a scenario where mitigation (e.g., social distancing) is
 successively imposed.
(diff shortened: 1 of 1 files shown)


jstac:
Thanks @longye-tian !  Much appreciated.  All changes are good.  Merging.

